PR #30, opened 2017-03-18
OpenMP Comparisons for mri_em_register

freesurfer-rge:
Scripts to compare OpenMP performance of mri_em_register when more than one copy is run at a time. The basic script to run is run_compare, which in turn calls run_one and run_two to run one or two copies of mri_em_register, and time them.


freesurfer-rge:
This is pending a bug fix for the output naming - please do not merge this PR.

PR #39, opened 2017-03-27
Bf samseg

ahoopes:


zkaufman:
Incorporating these changes in with the lcn_branch PR

PR #52, opened 2017-04-18
Added a stub file for testing purposes.

ahoopes:


ahoopes:
Just a stub file for testing git annex

PR #84, opened 2017-05-30
ENH: cache apt and ccache

yarikoptic:
It might speed up the build a bit thus giving more time for testing

might need few "runs" to see if there is an effect from ccache, so I will follow up this commit with subsequent "dummy" one to see if there is an effect on build time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ compiler: gcc
 
 os: linux
 
+cache:
+ - apt
+ - ccache
+
 addons:
   apt:
     packages:


yarikoptic:
ok -- use of ccache and apt cache drops time from 13 to 9 minutes.  imho worthwhile improvement ;)

zkaufman:
I see this change passed all tests and reduced build time. Can you provide brief description of what the changes to .travis.yml were before I accept PR?

yarikoptic:
First one is to cache Debian packages you get through apt. Second one is to make use ccache (c cache, avoids rebuilding sources if neither they not libraries they use had changed) and cache it's cache. Handy for use even locally during developing

PR #87, opened 2017-05-31
WiP NF:  make check testing in travis

yarikoptic:
sits on top of #84 

just a blind shot first -- I guess needs some additional devel pkgs and probably some data files to be fetched first.


yarikoptic:
hm, that is strange -- travis didn't pick up this last commit for testing... will push an amended commit to see if it helps

edit 1: seems to be a generic problem with travis/github ATM

glut/Makefile.am
```diff
@@ -67,8 +67,7 @@ endif
 
 test_glut_SOURCES=test_glut.c
 test_glut_LDADD=$(addprefix $(top_builddir)/, $(LIBS_MGH)) $(LIBS_VXL) -lz \
-	$(LIB_JPEG) $(LIB_TIFF)	$(X_LIBS) $(LIBS_OPENGL) $(top_builddir)/glut/libglut.a
-
+	$(LIB_JPEG) $(LIB_TIFF)	$(top_builddir)/glut/libglut.a $(X_LIBS) $(LIBS_OPENGL)
```
yarikoptic, line 70: FWIW  that is what I thought to try! ;)

zkaufman: I get two errors:

1) The job exceeded the maximum time limit for jobs, and has been terminated.

2) No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

I think I can fix (2). And I might be able to fix (1). But I'm confused, how does it know to split into 2 groups of tests. I see you added matrix, with 2 entries, but those list are incomplete. The tests it fails on isnt even included in that list - so Im not sure its even recognizing it.

zkaufman: Ok it looks like its ignoring the matrix and the complete set of tests is simply running in parallel. 'make check' will surely exceed the 50min allotment allowed by travis for a job:

  https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts

Either we need to get the matrix of tests working, or only a small subset of freesurfer tests can be integrated into the travis build.

yarikoptic: yeap -- as I have said somewhere -- it ignores 'TESTS' variable as far as I see it

my question to you (since I haven't trried yet locally, will do now) -- could you limit anyhow which tests get executed while doing it just straight in the shell

zkaufman: Many of the test scripts check for an environment variable which, if defined, will skip the test. For example, SKIP_MRI_CA_LABEL_TEST, SKIP_MR_CA_REGISTER_TEST, (and others) will skip those respective tests:

https://github.com/freesurfer/freesurfer/blob/dev/mri_ca_label/test_mri_ca_label

I suppose this could be generalized in some way to produce a subset of tests. But am I correct that we are limited to only 50 minutes. Perhaps some trickery can be used to get around this.

yarikoptic:
woohoo @zkaufman -- check it out -- it "works".  so it is a matter of timing all the tests now and then select wisely  ;-)

yarikoptic:
run on your box smth like
```shell
TESTS=$(find -maxdepth 1 -type d | sed -e 's,\./,,g' | grep -v '^\.' | tr '\n' ' ') ./run_selected_tests  2>&1 | tee run_selected_tests-output.log
```

yarikoptic:
@zkaufman is there a way to assess % code coverage from running some sub-set of tests?

zkaufman:
No their is not. We currently have no % code coverage analysis tools.

zkaufman:
@yarikoptic Woohoo! Integrated testing.
One question I have before accepting the PR... Is the fact that the entire build process gets completes for all TESTS entries expected behavior? It appears the entire build happens for all four TESTS entries in the matrix. I would think that only need happen once.

yarikoptic:
if I have got the question right -- it indeed does build freesurfer 4 times (once for each entry in the matrix) since those are independent from each other (besides reusing the same apt and ccaches from previous builds).  That is just how it is.  Better to build 4 times than none ;)  and build would succeed if all 4 build and test correctly ;)

PR #105, opened 2017-07-04
Enabled ex vivo scan processing -- added possibility to specify which…

leempko:
… brain structures are missing, build tailor-made affine registration atlas etc


ahoopes:
these changes have been moved and accepted in a new pull request

PR #136, opened 2017-09-11
Changed the template element in the L2S structure to mri_template bec…

dngreve:
…ause template conflicts with something in qdec


m-reuter:
I think it conflicts with the g++ who thinks that "template" is a key word. It is probably just included in QDEC for the first time, so that the error shows up there first.

PR #152, opened 2017-09-28
Bf gcc7.2.0 complaint fix

emtpgeek:
The function returns a pointer, not a boolean.   Prior to this fix, this specific return returned false rather than NULL which has the same 'zero' value but gcc 7.2 is complaining about it.

### DIFF
--- a/freeview/RenderView3D.cpp
+++ b/freeview/RenderView3D.cpp
@@ -1108,7 +1108,7 @@ SurfaceROI* RenderView3D::InitializeSurfaceROI( int posX, int posY )
 
   if ( !surf )
   {
-    return false;
+    return NULL;
   }
 
   lc_surf->SetActiveLayer( surf );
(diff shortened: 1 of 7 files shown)


m-reuter:
In my opinion this pull request has lots of improvements in a single pull. Some parts can easily be pulled (e.g. obvious fixes and added functionality, such as the surface proximity new code). I am not very happy with all the ifdef blocks. New code should be callable via command line flags (unless of course it is a bug fix).

emtpgeek:
If you would like me to change it to be a set of separate requests for small unconditional changes, I will be more than happy to do so.

On Sep 28, 2017 8:13 PM, Martin Reuter <notifications@github.com> wrote:

In my opinion this pull request has lots of improvements in a single pull. Some parts can easily be pulled (e.g. obvious fixes and added functionality, such as the surface proximity new code). I am not very happy with all the ifdef blocks. New code should be callable via command line flags (unless of course it is a bug fix).

emtpgeek:
I appear to have messed up this pull request - it was supposed to be a simple one line change independent of the other performance changes

I will try to understand what I have got wrong in my git interactions

PR #154, opened 2017-09-28
Remove the conditionals

emtpgeek:
Since the macro was defined at the top of the file, this is just eliminating dead code

### DIFF
--- a/mris_diff/mris_diff.c
+++ b/mris_diff/mris_diff.c
@@ -1,5 +1,3 @@
-#define BEVIN_IMPLEMENTATION
-
 /**
  * @file  mris_diff.c
  * @brief Compare two surfaces.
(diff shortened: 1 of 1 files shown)


ahoopes:
Is there a replacement for the --thresh option? Some of our regression tests are failing without it

emtpgeek:
Interesting.

-threshold was the way of specifying the range of acceptable differences.
Try just removing its use of the option – then it should pass.  Or I can modify the code to accept and ignore this option.

I replaced this with a hard coded set of tables in the new code, which may need to be adjusted slightly if just removing the option still has the test failing.

If you don’t want to remove the option,  this is the code to put into mris_diff.c, or I can do it…


   } else if (!strcasecmp(option, "--aparc2")) {

      if (nargc < 1) CMDargNErr(option,1);

      aparc2name = pargv[0];

      CheckAParc=1;

      nargsused = 1;

    } else if (!strcasecmp(option, "--thresh")) {

      if (nargc < 1) CMDargNErr(option,1);

      nargsused = 1;

    }

    else if (!strcasecmp(option, "--maxerrs")) {

      if (nargc < 1) CMDargNErr(option,1);

      sscanf(pargv[0],"%d",&MAX_NUM_ERRORS);

      nargsused = 1;

    }


/Bevin

ahoopes:
Okay, I'll re-add `--thresh` as an ignored option for now, and then remove it once I edit the tests

PR #159, opened 2017-10-08
Some performance improvements

emtpgeek:
Remove the conditionals, since this code will pass with the pulled mris_diff

There are two basic ideas enabled here.   

1: normalize a vector once to simplify its repeated usage.  

2:  calculate the angle using a quadratic approximation for arccos(x) when x is very close to 1.0, which is almost always the case for our usage.

### DIFF
--- a/include/matrix.h
+++ b/include/matrix.h
@@ -52,12 +52,10 @@ typedef struct
 MATRIX, VECTOR ;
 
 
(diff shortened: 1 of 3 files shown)


emtpgeek:
Oh, there was a third

3)  Use a simpler 3D vector in this performance critical code

PR #160, opened 2017-10-09
An alternative implementation of MRISaverageGradients 

emtpgeek:
This implementation is approx 8x faster resulting in an overall speedup of mris_register of about 2x.

It does this by moving the ripflag tests out of the loop and getting just the relevant data into a temporary secondary data structure that is much denser and hence makes much better use of cache and memory traffic.

Since this function takes up almost all of mris_register's time, it is worth this level of optimization, since this brings my timing run down from 0.25 hours to 0.13 hours

### DIFF
--- a/utils/mrisurf.c
+++ b/utils/mrisurf.c
@@ -21,6 +21,8 @@
  *
  */
 
+#define BEVIN_MRISURF
+
 #include <stdio.h>
 #include <string.h>
 #include <ctype.h>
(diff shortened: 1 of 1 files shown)


chrisadamsonmcri (commented):
Great work.
Small change. A 1-left bit shift could replace the multiplication on line 10942.

emtpgeek:
You are right, but I expect that an optimizing compiler will replace
    limit *= 2
with either
    limit += limit
or
   limit <<= 1
which ever, depending on the machine, faster.   

Thank you for reading through the change and noticing this.

emtpgeek:
Now I have to figure out why the checks failed...

emtpgeek:
Looks like a build system problem

"GPG error: http://repo.mongodb.org/apt/ubuntu trusty/mongodb-org/3.2 Release: The following signatures were invalid: KEYEXPIRED 1507497109"

ahoopes:
Fixed now. I'll accept the pull after tests of the current repo finish up

chrisadamsonmcri:
Good point. After reading on this the compiler should optimise with a bit-shift for a times 2 multiplication.

PR #164, opened 2017-10-09
run make from yml

ahoopes:


m-reuter:
this makes the log overflow, which is bad if you want to see the final lines and the test results. 
The || travis_terminate 1 causes the travis script to stop, if the build fails. This is good to skip the tests, but bad because we do not see the tail of the log file. Maybe you can print the tail of the log from within the make_script and then return from the script with the return value of the make command.

ahoopes:
Right - I'm just doing this to debug. Right now, every build fails - I think due to an expired GPG key on travis's side. Based off the tests I've done so far, make seems fails if the output is redirected but passes if it's not.

A few other people have been experiencing similar problems in the past day - https://github.com/travis-ci/travis-ci/issues/8554

PR #184, opened 2017-10-25
Nf bevin conditionalize minc

emtpgeek:
This change allows a build that does not need the MINC component at all.  This is desirable to Corticometrics because it allows the use of the code they need without this component being included, which means it doesn't have to be ported to some build configurations (other compilers, for instance).

This edit consists of several kinds of minor but wide ranging changes

(1) "Real" is replaced by "double" almost everywhere.   There are currently 3 typedefs of Real in the various header files, all defining it as "double", and unrelated to each other.  Furthermore a variety of the uses were not "typesafe" - Real was used when double was meant and vice versa

(2) "print" was being used instead of "printf" in several places, just because it is interchangeable

(3) "errno.h" was being included indirectly via volume_io.h, and needed to be made explicit

(4) Lastly the talairach transforms were being used but are almost always one of several options.  These were conditionalized out.

This edit does not add any new code or change any functionality when the conditionalization is not triggered

### DIFF
--- a/include/minc_volume_io.h
+++ b/include/minc_volume_io.h
@@ -26,6 +26,12 @@
 #ifndef MINC_VOLUME_IO_H
 #define MINC_VOLUME_IO_H
 
+#if defined(BEVIN_EXCLUDE_MINC)
+
+
+#else
+
+
 /*
  * Wrapper for MNI's volume_io.h, which has some annoyances which
  * must be circumvented.
@@ -52,9 +58,13 @@
 #ifdef Windows_NT
 #undef ERROR
 #endif // Windows_NT
+
 #include <volume_io.h> //from MNI
+
 /* remove unwanted warnings between hips_basic.h vs. volume_io/basic.h */
 #undef ABS
 #undef SIGN
+
+#endif
 
 #endif // MINC_VOLUME_IO_H
--- a/mri_cc_medial_axis/mri_cc_medial_axis.c
+++ b/mri_cc_medial_axis/mri_cc_medial_axis.c
@@ -90,9 +90,9 @@ typedef struct medial_axis_type_ {
 atom_type, MEDATOM ;
 
 
-static Real cc_tal_x = 64 ;
-static Real cc_tal_y = 55 ;
-static Real cc_tal_z = 64 ;
+static double cc_tal_x = 64 ;
+static double cc_tal_y = 55 ;
+static double cc_tal_z = 64 ;
 static LTA *lta = 0;
 static int rotatevolume();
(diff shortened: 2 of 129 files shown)


mri_cc_medial_axis/mri_cc_medial_axis.c
```diff
@@ -55,7 +55,7 @@
 #include "cma.h"
 #include "version.h"
 #include "error.h"
-#include "volume_io/geom_structs.h"
+//#include "volume_io/geom_structs.h"
```
nschmans, line 58: should this line be removed instead of commented-out?

utils/test/mrishash/mrishash_demo_100_find_coverage.c
```diff
@@ -101,7 +101,7 @@ void CoverageTest() {
   int rslt;
   int  xsi, ysi, zsi, xvi, yvi, zvi;
   int intens, halfscansize, closest_vertnum, vno;
-  Real xs,  ys,  zs,  xv,  yv,  zv;  // WTF is Real?
```
nschmans, line 104: ha ha

include/minc_volume_io.h
```diff
@@ -52,9 +58,13 @@
 #ifdef Windows_NT
 #undef ERROR
 #endif // Windows_NT
+
 #include <volume_io.h> //from MNI
```
nschmans, line 62: It appears to be the case that the structs VIO_General_Transform and VIO_Transform are the only things used from volume_io.h.  I'm thinking it makes sense to just include these structs directly here in this file (maybe adapted a bit to remove the multiple embedded structs and extraneous junk), and just not include volume_io.h at all.  this would sever the dependency completely, and would mean all the BEVIN_EXCLUDE_MINC conditionals could be removed (ie, returned to the original code... or better yet, replace with comments somehow delineating that that code block used MINC... ie, i'd hate to lose the work identifying these areas, even though they are 'fixed' elsewhere).  Where the adapted structs are declared here in minc_volume_io.h, comments could be added to acknowledge the source of the adapted code (ie, the copyright notice at the top of geom_structs.h).

ahoopes: Nick, should I accept this pull or wait until this is sorted out?

nschmans:
yes, please accept it.  i tested that it doesnt break our build when not using the BEVIN_EXCLUDE_MINC define
n.

ahoopes:
I've sorted out the conflicts and merged these changes through a separate pull https://github.com/freesurfer/freesurfer/pull/188

PR #198, opened 2017-11-06
Nf bevin replace minc

emtpgeek:
Implements the minc functionality needed by freesurfer, except if - unbeknownst to us - transforms other than Linear Transforms appear in xfm files.  None of the xfm files I have contain non linear transforms.

Removes most of the conditionals I put in when no transforms were available

Passes all my tests with no changes in results


emtpgeek:
Looking over this, I notice that the "inverse.c" that I disappeared and the minc_inverse.c that I added are very unrelated.  

I don't know why the previous functionality in "inverse.c" seems to be unneeded.  Nick may be able to shed some light here.

PR #206, opened 2017-11-22
Provide an alternative to building with the minc 3rd party library

emtpgeek:
I have created this pull request so all the interested people can see the way this alternative works, and provide input on any aspect I should have done differently.   If it is acceptable to all, then it can be pulled, but if there are needed changes, I'll do those and come back here...

/Bevin

Here is what I have done
        
            Extracted netcdf from within the mni package and made it into its own package
            
            Created an minc_substitute project similar to the utils project,
            and placed sources in it that are largely extracted from the minc 1.5.1 sources
                The required copyright notices have been copied from those sources.
                MGH Copyright's have been added in addition to the copied ones.
                Unneeded code was either not moved or is conditionalized out.
                Some code was rewritten rather than copied, to avoid dragging more unnecessary code.
                The new code still keeps the original copyrights because I at least looked at the original.
            
            Almost all of the minc_substitute code is conditionalized by BEVIN_EXCLUDE_MINC 
            but there are small portions of code that compile even when minc is used
            so that some testing can be done comparing the substitute code with the minc code
            
            The config and makefiles are modified to include this netcdf and minc_substitute

        Here is how to use it
        
          to avoid typing this path repeatedly...
            export SOME_NAME_FOR_PKGS_DIR=/home/rheticus/centos6-x86_64-packages
                or where-ever you put yours
        
          Make a directory within the third party packages containing netcdf's headers and libraries
            mkdir -p ${SOME_NAME_FOR_PKGS_DIR}/netcdf/current/{include,lib}
            cp ${SOME_NAME_FOR_PKGS_DIR}/mni/current/include/netcdf.h ${SOME_NAME_FOR_PKGS_DIR}/netcdf/current/include/
            cp ${SOME_NAME_FOR_PKGS_DIR}/mni/current/lib/libnetcdf.*  ${SOME_NAME_FOR_PKGS_DIR}/netcdf/current/lib/
          Optionally move mni out of the way, so you can be sure it not being accidently used...
            mv ${SOME_NAME_FOR_PKGS_DIR}/mni{,~not_avail}
            
          Do an normal build, except for adding --without-mni-dir
            ./setup_configure |& tee setup_configure_cout_cerr.txt | tail
            
            ./configure \
                    --disable-GUI-build \
                    --without-mni-dir \
                    --with-pkgs-dir=...as before... \
                    --prefix=...as before...`
            make
            
          Do a .mnc specific file operation...
            mri_binarize/mri_binarize \
                --min 0 --i ${FIND THIS PATH BY FINDING THE FILE}/average_305_2_blur.mnc \
                --o output_average_305_2_blur-nominc.mnc
      
          Do a normal build
            make clean
            mv ${SOME_NAME_FOR_PKGS_DIR}/mni{~not_avail,}
            ./setup_configure |& tee setup_configure_cout_cerr.txt | tail
            ./configure \
                    --disable-GUI-build \
                    --with-pkgs-dir=...as before... \
                    --prefix=...as before...`
            make
            
          Do the same .mnc specific file operation...
            mri_binarize/mri_binarize \
                --min 0 --i ${FIND THIS PATH BY FINDING THE FILE}/average_305_2_blur.mnc \
                --o output_average_305_2_blur-minc.mnc
      
          Cksum the two files to see they are the same
            cksum output_average_305_2_blur-nominc.mnc output_average_305_2_blur-minc.mnc


nschmans:
Bevin, 

It's looking like we're getting close to removing the build dependency on the minc library.  This will benefit the community by eliminating the need to build/install the LCN custom minc package, which is a real hassle.  

Would it be possible for you or someone to run the following test?

recon-all -s bert -autorecon1

In this first stage, .mnc file reading and writing is exercised (I think).

Going forward, if it looks like the new minc_substitute is working fine, then the next step is to modify configure.in to remove the *build* reference to the external mni package entirely.  I believe the *install* Makefile.am may still need to retain a reference to the mni tools, as certain legacy binaries are used (mnitotal) when recon-all is run and some flags are specified, but those can now be copied from any of the latest minc installations available, not our special build, and possibly not needed at all.

N.

emtpgeek:
Started at Tue Nov 28 01:04:51 PST 2017 
Ended   at Tue Nov 28 08:52:49 PST 2017
#@#%# recon-all-run-time-hours 7.799
recon-all -s bert2 finished without error at Tue Nov 28 08:52:49 PST 2017
done

But I see it is currently failing - I need to determine what has changed and fix it

PR #243, opened 2018-01-11
Fix ROMP-caused build problems on centos 6

emtpgeek:
Change the ROMP definitions and move their uses so they are not sensitive to the compiler version being used, successful build now on Centos 6 with gcc 4.4.*


ahoopes:
Thanks Bevin - I'll start a centos6 build now

PR #244, opened 2018-01-12
added label fusion - Lilla

ahoopes:


m-reuter:
Scripts should have an extension , here LabelFusion.sh else it can be confused with an binary executable

ahoopes:
I just realized LabelFusion actually is a binary - I'll ask Lilla what it is and move it accordingly

PR #258, opened 2018-01-23
Do not overwrite posteriors in T2 mode with different user-provided s…

jeiglesias:
…uffixes


ahoopes:
this PR was reimplemented in https://github.com/freesurfer/freesurfer/pull/259 to clear up potential git annex issues

PR #271, opened 2018-01-29
test: re-enabling tests run for make check in the utils/test directory

nschmans:
Tests that used to be enabled in the utils/test directory have been re-enabled.  Not sure why they were commented-out, but they pass when run on a center machine running centos6.


nschmans:
giving up on this branch

PR #273, opened 2018-01-30
bf: fixed netcdf.h error in /include/gcamorphtestutils.hpp; cleaned-up configure.in logic handling the local minc and netcdf

nschmans:
bf: fixed netcdf.h error in /include/gcamorphtestutils.hpp; cleaned-up configure.in logic handling the local minc and netcdf; renamed BEVIN_EXCLUDE_MINC define to USE_LOCAL_MINC


emtpgeek:
Very happy to see these!

PR #276, opened 2018-01-30
rf: mods supporting build on Mac High Sierra

nschmans:
rf: mods supporting build on Mac High Sierra using Homebrew install of gcc 4.9. See https://surfer.nmr.mgh.harvard.edu/fswiki/freesurfer_mac_developers_page


ahoopes:
Do sierra builds not compile because of this? https://github.com/freesurfer/freesurfer/blob/0f95ed1ee86116728fc2fb1be491a59678df1ebc/include/timer.h#L47
I might have to create a `HAVE_CLOCK_GETTIME` macro or something like that for OSX < 10.12 because commenting this out breaks the current El Capitan build

PR #285, opened 2018-02-03
Nf fast mris make surfaces 2

emtpgeek:
Tests: recon-all bert gets no differences

Begin to simplify (delete #if 0 code, rename a few functions) the interactions between mrisurf and mrishash as I prepare to make the hash table able to support a parallel version of the main expansion loop in mris_make_surfaces

This will require a per-thread hash table, hence some of the clarifications in when and how the hash tables come into existence, and the removal of the hash table writing the centroid into the vertices


emtpgeek:
Since this one will be superseded by a later one, I will close this one

PR #287, opened 2018-02-04
Nf fast mris make surfaces 4

emtpgeek:
Test:  recon-all bert gets identical results and intermediate files
         mris_make_surface  the old and new algorithms get the same answers to all questions (can be configured to use both and compare while running)

Implement new versions of the algorithms in mrishash that no longer need to modify the vertices temporarily, so can be used in parallel


nschmans:
a recent run using, we think, the latest code as of feb 2 or so, had this runtime info:
```
tool                    r/lh    total   clock times
----------------        ----    --      ----------
mris_sphere -q          2/2m    2m      4:00 start
mris_fix_topology       7/12m   12m     4:14 end
mris_make_surfaces wh   5/4m    9m      4:19 end
mris_sphere             3/5m    10m     4:24 end
mris_register           8/8m    15m     4:32 end
mris_make_surfaces pi   9/10m   16m     4:42 end
mris_volmask            8m              4:42 start
mris_anatomical_stats   1m      1m
mri_segstats            2m      2m      4:54 end
```
so it looks like the mris_ binaries still collectively add-up the runtime.  mris_make_surfaces is the slowest, but mris_register/fix_topolopy/volmask are not far behind.

ahoopes:
@emtpgeek since this merge, the `mris_expand` test runs endlessly on all of our build machines. This can be replicated by running `make check` in the `mris_expand/` dir. I wasn't able to track down the full source of this, but it seems like the code hangs in the call to `MRISminimizeThicknessFunctional(mris, &thick_parms, 10.0)` from `MRISexpandSurface`

emtpgeek:
Interesting – I’ll look into it

/Bevin

PR #292, opened 2018-02-06
nf: added git-rcs-keywords git filter 

nschmans:
nf: added git-rcs-keywords git filter to expand the existing CVS-stye tags like dollarIddollar, dollarAuthordollar, dollarDatedollar, as found in most .c/h files


nschmans:
needs more work...

PR #302, opened 2018-02-12
Clean up netcdf, fix for external ITK50, add fix for C++11

melrobin:
Added a few fixes for external ITK 50; cleaned up the netcdf path (-I../netcdf_3_6_0_p1 is already set) and replaced sm_11 with sm_52 for more modern GPUs.  I can make a change for later models later.  

Some changes were made to build with C++11.

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -1016,7 +1016,7 @@ AC_ARG_ENABLE(fermi-gpu,
 AM_CONDITIONAL(HAVE_FERMI_GPU, test "x$ac_have_fermi_gpu" = "xyes")
 
 # setup nvcc flags (default is for tesla-class, --enable-fermi-gpu overrides)
-NVIDIA_ARCH="sm_11"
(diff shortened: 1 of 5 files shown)


melrobin:
The continuous integration is failing because of netcdf.h.  I do not think that the tests have this in the correct location.   The netcdf_3_6_0_p1 directory containing netcdf.h is in the base freesurfer directory and so the -I../netcdf_3_6_0_p1 should obviate the need for the netcdf_3_6_0_p1 prefix on this line #include "netcdf_3_6_0_p1/netcdf.h".

johnmaia:
Couldn't the GPU arch (sm_10, sm_52, sm_61, ...) be a configuration attribute instead of something hardcoded?

melrobin:
This is probably a pretty good idea.  All architectures could be included up to sm_70

PR #314, opened 2018-02-15
Fix a missing add to x

emtpgeek:
This typo is causing the mris_expand problem

I am stunned it is not causing other major problems


ahoopes:
thanks!

PR #327, opened 2018-02-26
Nf realm 4

emtpgeek:
Deletes a lot of dead code from mrisurf.c, rather than maintaining it

Improves the timing support in ROMP to enable timing nested scopes

Parallelizes the hot loops of mris_fix_topology.  This causes very slight changes in some of the brain measurements, because the old algorithms did not have an exact parallel equivalent.  The new algorithms get identical results independent of the number of threads.

Adds a much faster mechanism for finding the faces that are needed to fix a defect

This code cuts fix topology times by about 1/3rd.
OLD:  mris_fix_topology rh  0.3 hours 4 threads
NEW: mris_fix_topology rh  0.2 hours 4 threads

The latter is the "realm" support. 

I looked at many synonyms for a region, zone, section, sector, subdivision, partition, ... 
of a volume into subvolumes but searching the sources forthe other choices found them 
being used already.  This particular choice had only one hit, in one comment.

Summarized, a RealmTree data structure is populated just before fixing the defects.  
It is adjusted as vertexs are moved during defect repair, and discarded when no longer needed.

A RealmTree the xyz volume of the surface is recursively broken into 2x2x2 subvolumes, 
each vertex is listed in the deepest subvolume it fits in.  Each face is listed in the 
deepest subvolume that all its vertexs fit in.  

Prior to fixing the defects a RealmTree is built.  The cost of the build is therefore 
amortized over many uses.   

A Realm is a (xLo..xHi, ...) subset of the whole coordinate space.  It does not have to be
a specific node in the RealmTree.

To find all the faces that might intersect a realm, a walk of the RealmTree visits only those
RealmTree nodes that intersect the Realm.   Typically this produces a lisst that is about 
1/20th of the all faces, thus dramatically speeding up the loop that chooses which faces to 
consider.


emtpgeek:
The overview of the realm support is also in a comment in realm.h

emtpgeek:
Oh, I also added a test to mris_fix_topology

PR #335, opened 2018-03-01
RF: use multiline statements in travis to avoid duplication/boilerplating

yarikoptic:
Sorry, I was sniffing around and could not just go by.
I think this should work and should make it easier to read and thus follow/fix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,48 +50,64 @@ addons:
 
 before_install:
   # on osx use brew to install autotools and compiler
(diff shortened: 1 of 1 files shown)


yarikoptic:
another common approach is just to get those as scripts somewhere (e.g. under `ci/` folder) and just call them from .travis.yml thus making travis much leaner and to the point while allowing to reuse those scripts across CI solutions

ahoopes:
Great! thanks Yaroslav

PR #346, opened 2018-03-13
travis debugging

ahoopes:


m-reuter:
The problem is that the ftp server at mgh is not reachable, so it does not download the libraries, therefore it cannot unpack and cd into that folder, but it does a cd .. at the end, ending up one directory above where it should be. 

It does not make sense to attempt a compile without the libraries. It would be nice to have travis error out earlier.

m-reuter:
Maybe we should put :
`curl ....  || travis_terminate 1 ; ...`
to have it stop there. 
Strange is, why can the mac built reach the ftp server, but the linux build not?

ahoopes:
That's a good idea - I can add it in. And it seems to be completely random, with different builds failing each time. It takes 3 or 4 individual restarts for a successful build.

I wonder why there have been so many issues connecting to the server the past week? We've actually been working on getting rid of the packages tarball entirely and having external users just build any extra libs within the repository.

PR #359, opened 2018-03-21
Bf mri make surfaces nondet 1

emtpgeek:
Added tracing to make it easier to find non-determinism that doesn't show up usually in the output files
Fix three problems discovered
There will be another patch based on this one for the harder problem


emtpgeek:
Something has gone wrong with my merge and its testing = will reopen when fixed

emtpgeek:
Okay - lets try this again

PR #361, opened 2018-03-22
Bf mri make surfaces nondet 2

emtpgeek:
https://surfer.nmr.mgh.harvard.edu/fswiki/mrisAsynchronousTimeStep_optionalDxDyDzUpdate

While the mrisAsynchronousTimeStep_optionalDxDyDzUpdate algorithm almost always gave repeatable results, it sometimes did not.  During debugging it was found that vertex movement is bent as it passed near another vertex.  On rare occasions this caused the vertex to move into a different subvolume.  This violated the assumed independence of the order of processing the subvolumes.  Code was added to detect such attempts to leave the subvolume, and to defer the motion until the serial execution of such vertexs, by reassigning the vertex to the set that have to consider the whole volume - processed in the second pass.

The tracing inserted during Bf_mri_make_surfaces_nondet_1 now shows that mris_make_surfaces gets identical intermediate MRIS states as well as producing the same output.

This fix is built on top of Bf_mri_make_surfaces_nondet_1, which also fixed other causes of rare slight variations.

### DIFF
--- a/utils/mrishash.c
+++ b/utils/mrishash.c
@@ -1571,11 +1571,10 @@ static int mhtDoesTriangleVoxelListIntersect(
     int const zv = (*voxel)[2];
 
     //----------------------------------------------------------
-    // Get corresponding bucket from mht. This *should* always
-    // succeed, given that the same info was just used to put xv,yv,zv
-    // into voxlist as was used to put faces into mht buckets.
+    // Get corresponding bucket from mht, if any.
+    // There might not be...
     //----------------------------------------------------------
-    if (!existsBuckets2(mht,xv,yv)) return (0);
+    if (!existsBuckets2(mht,xv,yv)) continue;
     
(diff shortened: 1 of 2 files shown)


emtpgeek:
Looks like we need to set the MRIs make surface test data again. Sigh

On Mar 23, 2018 7:34 AM, Andrew Hoopes <notifications@github.com> wrote:

Merged #361<https://eur03.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Ffreesurfer%2Ffreesurfer%2Fpull%2F361&data=02%7C01%7C%7C92af8ac92d934007135008d590239b3d%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636573404960808028&sdata=PBZmwKlKHwnWJlCtN1XhfjO4T9o5olDTMXD9nO1L8qA%3D&reserved=0>.

PR #385, opened 2018-04-12
Nf: Display 3D morph type in mri_info.

mu40:


buildqa:
Server connection failures caused some tests to fail, but otherwise looked OK.

PR #386, opened 2018-04-13
[REF] Update 'packages' to standard interface

ltirrell:
- Creates a `vxl` dir within `packages` with similar `makefiles` and `build_local` scripts as other packages
- Explicitly builds the latest release of ITK (v4.13) instead of 5.0 (which is currently in alpha)
  - Note: references to `itk-5.0` (in various forms) have been altered to `itk-4.13`. This could be changed so that v4.13 is added in addition to v5.0 in `configure.in` 

Note that the latest release of vxl (v1.17) doesn't build with gcc v5.4 and cmake v3.11, so it is built from the latest commit.

Todo (from @nschmans )
- [ ] all the 3rd-party dependencies needed to build the publicly distributed freesurfer should have a directory in /packages 
- [ ] a build_local and makefile in each package subdir which does a git clone of the latest _stable_ version of the package -- need explicit commit for vxl?
- [ ] package `makefile`s should check if `src` exist, and not clone code if it does
- [ ] the configure.in file supports automatically detecting and using a build of a package if a build is found in the /packages dir.  this is already the case for /packages/vxl and /packages/itk
- [ ] a README file is created in /packages which describes this standard and how to use (where it is expected that a builder will run the makefiles
- [x] add a makefile in /packages which builds all the subdir, or can build individual packages (see commit 973aa73 for initial version)


pwighton:
tetgen uses a [GPL license](https://en.wikipedia.org/wiki/TetGen).  We had previously removed it for this reason.  Do we need it again

packages/vxl/makefile
```diff
@@ -16,7 +16,7 @@ RANLIB:=$(shell which ranlib)
 
 export MAKE CMAKE CC CXX AR RANLIB
 
-checkout_cmd:=$(GIT) clone git@github.com:vxl/vxl.git
+checkout_cmd:=$(GIT) clone https://github.com/vxl/vxl.git
```
pwighton: I suggest doing something like `git checkout 3b74e3f8098f82be353c9cb5dbff46706a424b75` after cloning this repo.  This will give you a specific version of the code (in this case, the last release) so things don't break when they update their repo.

ltirrell: definitely a good idea... just not sure which commit to use, they haven't updated their release branch in 5 years. maybe it'd be worth posting an issue to their account to see if they recommend a specific commit? otherwise, we could use a commit that is used to successfully build FS (currently haven't done this)

buildqa:
Test failures were from fail to connect. so will merge in change.

- rob

PR #387, opened 2018-04-14
Nf faster distance map update 3

emtpgeek:
Reduces mris_fix_topology times during recon-all from
        7.2m -> 5.5m
        8.6m -> 6.9m

Does so by a series of changes to how matrices are allocated described in
    https://surfer.nmr.mgh.harvard.edu/fswiki/MorphoOptimizationProject_BetterSerialCode_matrices

and some changes to how the serial code for computing the distance to the nearest face is implemented, partially described in
    https://surfer.nmr.mgh.harvard.edu/fswiki/MorphoOptimizationProject_improvingParallelism

It also includes a fix to the incorrect calculation of the sign of the distance described in the fsdev mail (see below), with an environment variable that can return to the old behavior.  It now calculates what I believe is the correct sign.  If Bruce or Florent disagree, we will change it to the behavior they recommend.

Changes to recon-all results are in the acceptable range, and the pull request includes an update to mris_fix_topology test's .REF data

> 
> This code is weird ? it calculates the smallest distance but uses the
> SIGN(sign) of the last value tested to set the distance?s sign.? This 
> seems wrong to me, and it suggests that the sign is irrelevant ? but 
> it does cause the mris_fix_topology_test to get somewhat different answers if I correct it.
> 
> ?
> 
> I am going to do a recon-all run to see what difference it makes there also.
> 
> ?
> 
> It seems likely to me that the SIGN is, in fact, irrelevant otherwise 
> we would have detected this bug before now.
> 
> ?
> 
> /Bevin

The code basically says

    distance = huge
    three repeats of
        if (sometimes) {
            sign = expression2
            distance = SIGN(sign) * MIN(fabs(distance), expression3)
        }

So imagine that sometimes is always true and
expression2 returns -1, -1, +1
expression3 returns   9,  1,   9

then 
repeat 1 : distance is set to  SIGN(-1) * MIN(HUGE, 9)   i.e. -9
repeat 2 : distance is set to  SIGN(-1)  * MIN(|-9|,1)      i.e. -1
repeat 3 : distance is set to  SIGN(+1) * MIN(|-1|,9)      i.e. +1

BUT IF the code had happened to do the repeats in a different order, say in reverse, it would have gone
repeat 1 : distance is set to  SIGN(+1) * MIN(HUGE, 9)   i.e. 9
repeat 2 : distance is set to  SIGN(-1)  * MIN(|+9|,1)      i.e. -1
repeat 3 : distance is set to  SIGN(-1) * MIN(|-1|,9)        i.e. -1

It is weird that the sign does not come from the same repeat that the magnitude comes from and it should not matter what order the three repeats are processed in.

There are two possibilities.  Either the sign does not matter, or it should come from the repeat that the magnitude comes from.  Since we are currently using basically a random sign, one is tempted to say it does not matter and not compute it, but maybe algorithmically it does matter and fixing this would give a better result.


buildqa:
Tests that failed merged because of connection failure, and not test error, so will merge in.

- rob

PR #391, opened 2018-04-16
don't build fsgd_wrap.c with non-GUI build (includes tcl.h)

nschmans:
- removed fsgd_wrap.c from non-GUI build because it includes tcl.h
- other minor improvements


buildqa:
Merging this in.  I noticed some errors with freesurfer compiling with gcc5.3 on Ubuntu 16.04, but I will try it again with these changes.

- rob

PR #394, opened 2018-04-19
Nf faster distance map update 5

emtpgeek:
Same results on recon-all as dev branch
Next significant decrease in the times to do mris_fix_topology

before:FSRUNTIME@ mris_fix_topology lh  0.0674 hours 4 threads
after   :FSRUNTIME@ mris_fix_topology lh  0.0519 hours 4 threads
before:FSRUNTIME@ mris_fix_topology rh  0.0871 hours 4 threads
after   :FSRUNTIME@ mris_fix_topology rh  0.0667 hours 4 threads

Achieved by (next step is to write these up in the Wiki)
1) Reducing the number of nodes in the realm tree by not making children until needed
2) Removing excess locking of the realm tree
3) Cached the psuedo norms rather than recomputing
4) Created per-thread distance temporary rather than locking a shared one
5) loop invariant'ed the k->z calculation
6) Added a fast test - described below - that eliminates many of the points from consideration

The k->z loop invariant:
Consider

    for i in 0..maxI
       for j in 0..maxJ
            v = f(j)

The f(j) returns the same value when i is 0, 1, 2, ...   but the compiler does not loop-invariant it from the inner loop because it depends on j.   The solution is to rewrite this as
   
    for j in 0..maxJ
            cache[j] = f(j)

    for i in ...
       for j in ...
            v = cache[j]

Now the expensive function f(j) is called maxJ times  rather than maxI*maxJ times

The fast test path:

The code is calculating the min distance from a point to any point on a triangle.  It now determines a fast crude lower bound on this distance, and ignores the point if the current min is less than this lower bound.   The lower bound is based on some math, namely

  //	    	let IJK be any point in the box, and FP be any point on the face
  //	    	    distance(IJK, Center) <= distance(IJK,FP) + distance(FP,Center) 	hence
  //	    	    distance(IJK, Center) - distance(FP,Center) <= distance(IJK,FP)

so when the LHS is larger than the current mininum distance for IJK, it is not worth further consideration of this face

distance(FP, Center) is itself less than the distance from center to furtherest vertex, which is loop invariant of the i,j,k loops

distance(IJK, Center)  is sqrt( square(x-cx) + ... )    The code exploits     sqrt(a) < sqrt(b)  if and only if   a<b   to eliminate the need to calculate this sqrt

The code gets a pointer to the (i,j,kmin..kmax) vector to avoid having to re-index every time around the fast test path loop

### DIFF
--- a/mris_fix_topology/test_mris_fix_topology_p_timing
+++ b/mris_fix_topology/test_mris_fix_topology_p_timing
@@ -1,0 +1,80 @@
+#!/bin/tcsh -f
+
+umask 002
+
+printenv | grep FREESURFER
+
+# 1 2 4 8
+foreach threads ( 1 )
+
+  # 2 -1
+  foreach niters ( 2 )
+    if ($niters != -1) then
+      set NITERS_OPTION = "-niters $niters"
+      set NITERS_EXTENSION =
+    else
+      set NITERS_OPTION = 
+      set NITERS_EXTENSION = _unlimited
+    endif
+  
+    #  currently available:
+    # 	.dev .nf_faster_distance_map_update_4 .nf_faster_distance_map_update_5
+    #
+    # Don't forget to...
+    #   cp mris_fix_topology{,.nf_faster_distance_map_update_5}
+    #
+    foreach branch ( .nf_faster_distance_map_update_5 )
+
+      set feature=0
+
+      set extension = -B${branch}-F${feature}-N${niters}-T${threads}
+
+      # the simple cmd
+      #
+      set cmd=(../mris_fix_topology${branch} \
+              $NITERS_OPTION \
+              -mgz \
+              -sphere qsphere.nofix \
+              -ga -seed 1234 bert lh)
+
+      # extract testing data
+      #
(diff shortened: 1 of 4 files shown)


buildqa:
Merging this in as test failures were from error in getting files from git-annex.  I have been using $FREESURFER_HOME/matlab_runtime (and also $MNE_ROOT/matlab_runtime) as the location for a minimal matlab runtime environment that does not require the Matlab application to be installed.  It can one downloaded from https://www.mathworks.com/products/compiler/matlab-runtime.html

emtpgeek:
Additions describing this work made to
https://surfer.nmr.mgh.harvard.edu/fswiki/MorphoOptimizationProject_improvingParallelism
https://surfer.nmr.mgh.harvard.edu/fswiki/MorphoOptimizationProject_BetterSerialCode
https://surfer.nmr.mgh.harvard.edu/fswiki/MorphoOptimizationProject_ReducingMemoryTraffic
https://surfer.nmr.mgh.harvard.edu/fswiki/MorphoOptimizationProject_distanceToNearestFace

PR #402, opened 2018-04-27
Fixing possible infinite recursion in trc/blood.cxx.

nagaflokhu:
FindCenterStreamline function (in trc/blood.cxx) calls itself with different checking options set to true or false in certain cases. The base case (else branch on line 2429) potentially calls itself to infinity if it happens that hdmin's value is not changed from numeric_limits<double>::infinity(). I presume this has to do with some characteristic of a streamline being unsatisfactory. In any case, FindCenterStreamline should not call itself again if all the check variables are already set to false.

I haven't looked in detail at what FindCenterStreamline does, so my fix may not be satisfactory, but I suspect this is something that should be fixed and that the fix is similar to my solution.


buildqa:
Merging since test failure was from network connection.
- rob

PR #404, opened 2018-04-28
Fix a typo that got the buffer size wrong, 

emtpgeek:
but was so large testing didn't find


buildqa:
-rob

PR #405, opened 2018-04-28
Nf faster distance map update 6

emtpgeek:
Reduces mris_fix_topology -mgz -sphere qsphere.nofix -ga -seed 1234 bert lh from 0.05h to 0.045h by
(a) deferring a sqrt, and (b) using schedule(guided) 

as described in https://surfer.nmr.mgh.harvard.edu/fswiki/MorphoOptimizationProject_improvingParallelism and https://surfer.nmr.mgh.harvard.edu/fswiki/MorphoOptimizationProject_BetterSerialCode

No changes in results


buildqa:
-rob

PR #406, opened 2018-04-29
Nf faster distance map update 7

emtpgeek:
This adds onto the "Nf faster distance map update 6" and the "bf_wrongBufferSize" pull requests

This further reduces the mris_fix_topology time for the lh, now down to 2.6mins (2.8 for update 6, 3.0 for the current dev stream)

It adds an optimized sort that can use threads and is optimized for sorting scalar types such as int, and uses it to speed up one of the previous calls to qsort that was happening in the critical serial portion of each iteration.

The sort is described in
 https://surfer.nmr.mgh.harvard.edu/fswiki/MorphoOptimizationProject_BetterSerialCode


buildqa:
-rob

PR #407, opened 2018-04-30
Nf faster distance map update 8

emtpgeek:
Various small tweaks along the hottest paths to reduce topology fixing from 2.6 mins to 2.3 mins

1 - change some sqrt to sqrtf 
2 - eliminate some bounds checking that was on the hot path
3 - use memset to reset the deferred face norm flags
4 - add a cache to computeDefectFaceNormals to speed up finding the unique faces
5 - reorganized defectMaximizeLikelihood as ..._new
6 - added updateDistanceEltFromSignArgAndSquareNoLockNeeded, which simplifies the hottest path and which starts the sqrtf well before it is needed, even if it is not needed, because the h/w can throw it away and the DIV unit is not busy doing anything else
7 - replaced some / 3.0 with *0.33333f in places where the precise accuracy of the double precision divide was not needed
8 - changed the way the 'passed' set is implemented in realm.c to speed up the emptying of the set.  Documented the approach in the Wiki.  Basically the set is a vector of ints, and a counter.  Elements are added to the set by setting set[element] = counter;  If set[element] == counter, the element is in the set.  The set is emptied by counter++.
9 - reorganized the finding of the ignorable k values to avoid branch mispredicts.

Note: There are other places where  A /= N, B /= N, ... could be changed into   A *= 1.0f/N, ...  but I did not do so because it changed the results.  The new results are probably equally valid but different, and the win was not, in my opinion, worth pursuing at this time

### DIFF
--- a/include/base.h
+++ b/include/base.h
@@ -89,8 +89,8 @@ extern "C" {
 
 // assertions
 //
-#define cheapAssert(TST)  { if (!(TST)) *(int*)-1 = 0; }
-#define costlyAssert(TST) { if (!(TST)) *(int*)-1 = 0; }
+#define cheapAssert(TST)        { if (!(TST)) *(int*)-1 = 0; }
+#define costlyAssert(TST) //    { if (!(TST)) *(int*)-1 = 0; }
 
 // Regardless of whether the __real_malloc etc. or the __wrap_ ones, it is still desirable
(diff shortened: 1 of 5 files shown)


buildqa:
Merging as failure was due to connection timeout.
- rob

PR #415, opened 2018-05-02
Nf icc build 1

emtpgeek:
Modify the configure script to make it easier to use the Intel compiler, and to make the Intel MKL optional when you are using it

Fix a bug that gcc did not notice but icc did - accidental sharing of variables being used for inner loop control but declared outside the parallel loop


buildqa:
Merging in as any test failures were from network, and a test build for CentOS 7 compiles.
- rob

PR #424, opened 2018-05-08
How did this not get found before?

emtpgeek:
Uninitialized member caused a crash when compiled with icc but,weirdly, not with GCC but it should have...

recon-all runs with identical results


emtpgeek:
Included in #425

PR #429, opened 2018-05-17
bf: travis

ahoopes:
Temporarily disable tests, since the annex fails and they're not really telling us anything important right now anyway. I've also extracted out a travis bash script for easier debugging of build faiures (since travis will continue running steps even if one fails)


yarikoptic-private:
I was unpleasantly surprised to find out that tests got completely disabled...
- what problem with git-annex did you experience? could you please point me to the failing builds? if anything fails with it I would be happy to help to get it fixed -- it must work
- `tests ... not really telling us anything important right now` - which is kinda a point of CI to know that nothing "important" was ruined.  I could only encourage to keep code base tested

ahoopes:
The travis checks had been failing for ~two months due to connection failures with our servers (for downloading the pre-built packages and the annex data). We've been in contact with travis support and finally got the curl command working recently for the packages data. However we weren't able to resolve the git-annex failures, so I disabled the tests temporarily in order to actually get some useful feedback from the travis runs instead of random failures.

Also, travis only runs about 20% of our full testing routine (working on expanding this). So most of the time - back when things were running smoothly - it seemed that travis would miss the test failures that were later caught by our nightly checks. Turning them off for now isn't really doing any harm (other than delaying failure notifications until the end of the day).

Here's an example of a [failing travis build](https://travis-ci.org/freesurfer/freesurfer/jobs/375612308). These annex errors seem to happen randomly (about 75% of the time), but if you think you could get it fixed that would be awesome - thanks!

yarikoptic:
great to hear that there is daily testing going on ;)  well - I will leave it up to you on either to resurrect testing

git-annex: 
- within the past 2 months there was a number of refactorings and fixes which should have made downloading more resilient to network/server misbehaviors , although I must say that I thought that the version you have been using in that failing one (6.20180416) should have got already some at least done... but the was some recent one also about (not) decompressing compressed streams which might be relevant since message was "verification of content failed" so download worked but check didn't
- just for the sake of making sure that problems with annex are fixed, would be kind to restart that elderly build: https://travis-ci.org/freesurfer/freesurfer/jobs/375612308  (there should be a button on top right to restart it -- I am not authorized to do that for this repo)?  thanks in advance
- edit:  if connectivity with your server remains problematic, I would be happy to mirror needed content from our server thus providing redundancy of sources for git-annex so it could download from either one

PR #430, opened 2018-05-17
ITK is not always available

emtpgeek:
This was causing a make failure on Ubuntu systems that did not have itk which in turn was causing a make install failure


ahoopes:
Fixed with https://github.com/freesurfer/freesurfer/pull/433

Issue #264, opened 2018-01-26
Starting Freeview


melrobin:
### Platform details:

- FreeSurfer version: latest GitHub version
- Operating system name and version:  Fedora 25

### Summary
I am able to build FreeSurfer from source, but when starting freeview I get the following error:
[melrobin@scorpion ~]$ freeview
This application failed to start because it could not find or load the Qt platform plugin "xcb"
in "".  

Reinstalling the application may fix this problem.
Abort (core dumped)

There should be a platforms directory somewhere, but it does not appear.
### Actual behavior
Error message as above
### Expected behavior
GUI to appear

melrobin:
Trying to do some troubleshooting:
[melrobin@scorpion ~]$ ldd $FREESURFER_HOME/bin/freeview.bin | grep xcb
        libxcb.so.1 => /lib64/libxcb.so.1 (0x00007efe4af38000)

So the library libxcb.so is there, but I thought libqxcb.so is needed

melrobin:
The Qt in the centos_x86-64 package directory does not have a plugins directory, but my system Qt does.

melrobin:
I was able to build Qt 5.6.2 and move the plugins directory in the qt subdirectory of packages and modify QT_QPA_PLATFORM_PLUGIN_PATH environment variable and was able to launch freeview.bin.  I'll try additional experiments, but it seems that plugins needs to be distributed in centos6-x86_64-packages

buildqa:
You could also try the following to check if there are any other missing lib references,

$ ldd $FREESURFER_HOME/bin/freeview.bin | grep "not found"

Since ldd calls the loader, then environment variables like LD_LIBRARY_PATH will affect the result, but QT specific environment variables will not, I believe.  (I'm not suggesting you set LD_LIBRARY_PATH though).  

If you are expecting to get your qt5 libs etc. from the system, then I would check that the package qt5-qtbase.x86_64 is installed, and maybe qt5-qtbase-devel.x86_64 (or that is what I recall from CentOS).

You can also try to see where the qtlibs are coming from via
$ ldd $FREESURFER_HOME/bin/freeview.bin | grep -i libqt

melrobin:
Check out the contents of the qt under the packages tarball.  It does not contain a plugins directory and so the directory will not contain libqxcb.so.

mpantoja314:
I have the same problem, I can compile freesurfer from scratch and everything seam to be working but I can't run FreeView. Same problem with "xcb"
[...]

PR #444, opened 2018-06-14
fix Qt platform plugin xcb missing in freeview 

m-reuter:
assuming plugin directory is in package tar ball (please put it there, with all plugins needed by freeview). Once plugins are also distributed, this should solve the issue #264.


irekpo:
plugins are still missing in freesurfer-linux-centos7_x86_64-dev.tar.gz

PR #460, opened 2018-07-19
WIP: Infant recon for non-MGH users

larsoner:
Trying to run the `infant_recon_all`. WIP because I haven't completed a run, but this is to keep track of the changes that were necessary. @lillazollei let me know if this is helpful or not!

1. Need to add to the Wiki that ~~MATLAB~~ the [MATLAB R2012b Runtime](https://www.mathworks.com/products/compiler/matlab-runtime.html) is required, and the `MATLAB_ROOT` env var needs to be set to point to this.
2. Update wiki that FSL is required.
3. Update wiki that [dramms](https://www.nitrc.org/projects/dramms/) is required.
4. ~~Packaging needs to be fixed to include some or all of `scripts/InfantFS`, or these files need to be moved to `bin` during packaging (if the latter, `InfantFS` instances will need to be replaced in the scripts).~~
5. Use `LabelFusion` from GitHub master rather than the archive.
6. Do not use a relative path for the skull-stripped input or it will break in the `norm.nii.gz` linking step.

We'll see if it runs all the way through now.

### DIFF
--- a/scripts/InfantFS/SetLabelFusionParams.csh
+++ b/scripts/InfantFS/SetLabelFusionParams.csh
@@ -1,8 +1,8 @@
-#!/bin/tcsh -ef    
+#!/bin/tcsh -ef
 
 #setenv CMDDIR `pwd`
-setenv CMDDIR $FREESURFER_HOME/scripts/LabelFusionBetterDist/
-setenv MLAB /autofs/cluster/matlab/8.0/
+setenv CMDDIR $FREESURFER_HOME/scripts/InfantFS/
+setenv MLAB $MATLAB_ROOT
 setenv MRFparams "1 5"
 
 setenv LFOPTIONS1 "2 41 3 42" # new -- cerebral WM hack
--- a/scripts/InfantFS/SetLabelFusionParams.csh
+++ b/scripts/InfantFS/SetLabelFusionParams.csh
@@ -1,7 +1,7 @@
 #!/bin/tcsh -ef
 
 #setenv CMDDIR `pwd`
-setenv CMDDIR $FREESURFER_HOME/scripts/InfantFS/
+setenv CMDDIR $FREESURFER_HOME/bin/
 setenv MLAB $MATLAB_ROOT
 setenv MRFparams "1 5"
 
--- a/scripts/InfantFS/infant_recon_all
+++ b/scripts/InfantFS/infant_recon_all
@@ -37,7 +37,7 @@ set setksize            = 0
 
 set useMI               = 0
 
-source set_infant_recon_params.csh
+source $FREESURFER_HOME/bin/set_infant_recon_params.csh
 set TEMPLATE_SUBJECTS = ($CNYBCH_SUBJECTS) # the full CNYBCH training set
 set TEMPLATE_AGES     = ($CNYBCH_AGES)
 setenv TEMPLATE_SUBJECTS_DIR  $CNYBCH_TEMPLATE_SUBJECTS_DIR 
--- a/scripts/InfantFS/set_babydev_packages.csh
+++ b/scripts/InfantFS/set_babydev_packages.csh
@@ -3,7 +3,7 @@
 ##### scripts location
 
 set FS_HOME = $FREESURFER_HOME
-setenv FSSCRIPTSDIR      $FS_HOME/scripts/InfantFS/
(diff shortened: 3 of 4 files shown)


larsoner:
Actually I think everything is packaged fine as long as the paths are set to what they will be once things are packaged, i.e. `freesurfer/bin` instead of `freesurfer/scripts/InfantFS` (the `scripts` dir does not exist in the packaged verison). Changes have been pushed to reflect this. We'll see how far I get.

larsoner:
FYI now the error I get in the script can be replicated locally with:
```
larsoner@bunk:~/applications/freesurfer-infant/bin$ ./run_LabelFusion.sh /usr/local/MATLAB/MATLAB_Compiler_Runtime/v80/ --help
------------------------------------------
Setting up environment variables
---
LD_LIBRARY_PATH is .:/usr/local/MATLAB/MATLAB_Compiler_Runtime/v80//runtime/glnxa64:/usr/local/MATLAB/MATLAB_Compiler_Runtime/v80//bin/glnxa64:/usr/local/MATLAB/MATLAB_Compiler_Runtime/v80//sys/os/glnxa64:/usr/local/MATLAB/MATLAB_Compiler_Runtime/v80//sys/java/jre/glnxa64/jre/lib/amd64/native_threads:/usr/local/MATLAB/MATLAB_Compiler_Runtime/v80//sys/java/jre/glnxa64/jre/lib/amd64/server:/usr/local/MATLAB/MATLAB_Compiler_Runtime/v80//sys/java/jre/glnxa64/jre/lib/amd64/client:/usr/local/MATLAB/MATLAB_Compiler_Runtime/v80//sys/java/jre/glnxa64/jre/lib/amd64

Error:Cannot find CTF archive /home/larsoner/applications/freesurfer-infant/bin/LabelFusion.ctf
```

larsoner:
From the FreeSurfer list I see this can be due to a bad binary, so I replaced the one from the archive with the one from GitHub and now it's running without error (at the moment).

ahoopes:
Oh okay - thanks for pointing this out. This usually happens when a compiled matlab binary gets stripped. I'll fix this when I update the new infant distributions. Are there any more issues with the command paths?

larsoner:
So far this has fixed the path issues, yeah. I got one subject to run to completion with these changes, which is great! But it didn't do any surface generation because there was no GM/WM boundary (re-running it now to force it to use a template to use for this), so I probably haven't hit complete coverage yet. I'll update if there are other problems :)

FWIW in the future there are some issues that could be fixed in the scripts themselves, specifically:

1. `.mgz` is not supported as an as arg to `--masked` (`.nii.gz` at least works, which can be made with `mri_convert`), and
2. passing a relative path to `--masked` rather than an absolute one breaks during a `norm.nii.gz` symbolic linking.

And as I wrote above, the wiki / instructional email probably needs some updates. But these are not show-stoppers.

lillazollei:
Thanks, Eric. I am in the process of updating these.
Lilla

On Thu, 19 Jul 2018, Eric Larson wrote:

> 
>         External Email - Use Caution        
> 
> So far this has fixed the path issues, yeah. I got one subject to run to completion with these changes, which is great! But it didn't do any surface generation because there was no GM/WM boundary
> (re-running it now to force it to use a template to use for this), so I probably haven't hit complete coverage yet. I'll update if there are other problems :)
> 
> FWIW in the future there are some issues that could be fixed in the scripts themselves, specifically:
>
>  1. .mgz is not supported as an as arg to --masked (.nii.gz at least works, which can be made with mri_convert), and
>  2. passing a relative path to --masked rather than an absolute one breaks during a norm.nii.gz symbolic linking.
> 
> And as I wrote above, the wiki / instructional email probably needs some updates. But these are not show-stoppers.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.[AGLBoagb_4ohnmzSuhIDC2_ergfjsA06ks5uIMoIgaJpZM4VWcsS.gif]
> 
> 
>


The information in this e-mail is intended only for the person to whom it is
addressed. If you believe this e-mail was sent to you in error and the e-mail
contains patient information, please contact the Partners Compliance HelpLine at
http://www.partners.org/complianceline . If the e-mail was sent to you in error
but does not contain patient information, please contact the sender and properly
dispose of the e-mail.

PR #467, opened 2018-07-24
Bf nan segv

emtpgeek:
The search went beyond the grid, which caused an index-out-of-range, which corrupted another data structure turning some floats into nan's


ahoopes:
thanks!

PR #490, opened 2018-08-13
Add labels for lobe parcellations to LUT

ltirrell:
This adds label values to the LUT for each lobar parcellation obtained from `mri_annotation2label --lobesStrict`, following the color convention used for the `wm-hemi-lobe` lobes in the lines below (eg the frontal-lobe value is the same as superiorfrontal, etc).

### DIFF
--- a/distribution/FreeSurferColorLUT.txt
+++ b/distribution/FreeSurferColorLUT.txt
@@ -613,9 +613,29 @@
 4034    wm-rh-transversetemporal            105 105 55  0
 4035    wm-rh-insula                        20  220 160 0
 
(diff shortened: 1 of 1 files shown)


ltirrell:
Note that `--base-offset 300` was used instead of 200 for the wm parcellations, since the 1200 values were already taken by other labels.

ltirrell:
On second thought, the color values looked bad when following the convention of the `wm-hemi-lobes`... I changed them back to what they would be if they were still labeled as 1001-6/2001-6, but changed insula to match its usual value.   I am open to changing these values if you'd prefer something different.

PR #505, opened 2018-08-24
Cmake minimal

ahoopes:
This the first step toward implementing build components (or modules) by allowing a *minimal* build that only compiles and installs the files necessary for the standard recon-all stream. This can be configured on the command line with `-DMINIMAL=ON`. This setting emulates the "small dist" autotools build option. In the future, as modules are added, this should maybe be switched to something like `-DALL_MODULES=OFF` or `-DDEFAULT_MODULES=OFF`


nschmans:
Thanks for adding this!   the build/install works for me on the Mac.

 I'm wondering if a build option supporting this could be added to CMakeLists.  I have a bit of an aversion to adding defines to the cmake command line, and a preference for using ccmake and toggling build options.   I can add it if you want.

ahoopes:
Sure! https://github.com/freesurfer/freesurfer/commit/3c89a6bc23ddfd6cf74efe9f30ab4718ead6c678

PR #513, opened 2018-08-30
Fixed ITK_DIR on macOS

mu40:
When compiling FS on macOS, prebuild 3rd-party libraries available from "http://surfer.nmr.mgh.harvard.edu/pub/data/fspackages/prebuilt/osx10.11-packages.tar.gz" are needed.

Currently, this package contains ITK 4.13.0, whereas version 5.0.0 is expected within CMakeLists.txt. Until someone makes a newer version of the package (and updates the URL in CMakeLists.txt), the correct ITK_DIR should be set if on macOS. See attached configuration attempts with CMake.
[problem.txt](https://github.com/freesurfer/freesurfer/files/2336980/problem.txt)
[solution.txt](https://github.com/freesurfer/freesurfer/files/2336981/solution.txt)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,11 @@ find_package(PETSC)
 
 # -------- itk --------
 if(NOT ITK_DIR)
-  set(ITK_DIR ${FS_PACKAGES_DIR}/itk/5.0.0)
+  if(APPLE)
+    set(ITK_DIR ${FS_PACKAGES_DIR}/itk/4.13.0)
+  else()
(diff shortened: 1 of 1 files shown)


ahoopes:
Hey Malte, the PR I have open now resolves this. So should be fixed soon

mu40:
Hi Andrew, sounds good. I found another issue: compilation fails at some point if the C++ standard is not explicitly set to C++11. Do you want me to submit a PR for this, or does yours cover this as well?

ahoopes:
That should be covered as well. I set `CMAKE_CXX_STANDARD=11` globally

mu40:
That will do, thanks.

PR #518, opened 2018-09-04
Nf mrisurf reorg c

emtpgeek:
This is the first cut at the splitting of mrisurf down from 80,000 lines into a set of smaller files with a subsystem structure

The internals of these files need a lot more reorg, but by checking them in now, Bruce and others will be able to continue to make changes to them without me having to do a complicated merge afterwards.


emtpgeek:
Missing files - working on it

PR #538, opened 2018-09-23
Nf topo 2

emtpgeek:
This checkin enables the separation of the vertex[] fields from the vertex_topology[] fields.  This may show up in code that is not built by me.  Typically 

    VERTEX* const v = mris->vertices[vno];
    v->vnum, v->f and some others

have to be changed to 

    VERTEX_TOPOLOGY const * const vt = mris->vertices_topology[vno];
    VERTEX                                 * const v  = mris->vertices               [vno];
    vt->vnum, v->f, and some others

Almost all the changes here come from renaming MRISsetNeighborhoodSize to MRISsetNeighborhoodSizeAndDist because that is what it does.

The rest of the changes are adding some consistency checking,  tidying up some of the MRIS construction and destruction, and adding some additional but as yet unused fields to mrisurf.h so that I won't have to add them in the next round - I am trying to make changes be either (a) wide-spread but trivial, or (b) one very specific non-trivial step.

I had to add some more diagnostic support to find the tiny difference that was introduced when my new code didn't set mris->nsize to 1 when constructing the MRIS.  This code will, I'm sure, be useful for tracking down more tiny changes in the future.

Gets identical results to recon-all.


rpwang17:
This change appears to have broken a lot of builds for me, including freeview.

ahoopes:
I just noticed that - looking into it now

rpwang17:
It looks like we need to use VERTEX_TOPOLOGY to access some of the data members that used to be in VERTEX. A TON of codes will need to be changed if that's the case!

ahoopes:
It doesn't require quite as much updating as expected. I'm about halfway done with a complete fix

rpwang17:
I can fix the freeview part.

emtpgeek:
I appreciate you both doing this   I know how tedious it was to do the portion of the code my system does compile.

I was expecting to see a lot of compilation errors from the pull request build.

Is there a way for me to use the pull request systems to do a complete build and complete test?

On Sep 25, 2018 5:20 AM, rpwang17 <notifications@github.com> wrote:

I can fix the freeview part.

PR #547, opened 2018-09-30
Nf topo 8

emtpgeek:


emtpgeek:
DONT PULL THIS YET - I am putting this here so you can see what it involves, and so we can resolve any remaining conflicts or issues in the GUI code 

This request builds on top of nf_topo_7_merged_dev, so we need to do that one first, and then I will merge dev into this one.

This 

1) reorganizes mrisurf_topology.c into a much more coherent order, 

2) moves some of the mark stuff over the mrisurf_val.c, 

3) uses the same code to compute wide neighborhoods as it uses to compute the v3num ones, and 

4) has an _old and _new implementation of the code that completes the topology.  The _old code is still being used, because the _new code results in edges being in an equally valid but different order that results in slight differences during recon-all

emtpgeek:
Superseeded by nf_topo_8_merged_dev

PR #559, opened 2018-10-10
Update build_packages script

### LINKED ISSUE #553: numpy and nibabel are imported when running packages/build_packages.py
### Platform details:

- FreeSurfer version: current dev (commit 2194c58)
- Operating system name and version: ubuntu 16 docker container

### Summary
The `packages/build_packages.py` script requires `numpy` and `nibabel` due to an `import *` within the the `python/freesurfer/__init__.py` file.

### Actual behavior
```
root@e64f53182357:/code/packages# ./build_packages.py --no-tetgen "${CODE_DIR}/packages"
Traceback (most recent call last):
  File "./build_packages.py", line 33, in <module>
    from freesurfer.log import *
  File "/code/python/freesurfer/__init__.py", line 3, in <module>
    from ._surface import *
  File "/code/python/freesurfer/_surface.py", line 1, in <module>
    import numpy as np
ImportError: No module named numpy
```
(after `pip install numpy`)
```
root@e64f53182357:/code/packages# ./build_packages.py --no-tetgen "${CODE_DIR}/packages"
Traceback (most recent call last):
  File "./build_packages.py", line 33, in <module>
    from freesurfer.log import *
  File "/code/python/freesurfer/__init__.py", line 5, in <module>
    from .freeview import fv
  File "/code/python/freesurfer/freeview.py", line 4, in <module>
    import nibabel as nib
ImportError: No module named nibabel
```
### Expected behavior
Ideally, no 3rd party modules should be required for the build script

### How to replicate the behavior
Use a python environment without `numpy` or `nibabel` installed, then run the `pacakges/build_packages.py` script, as described under "Making the third party packages" on the [wiki](https://surfer.nmr.mgh.harvard.edu/fswiki/CMake).

### Details
I'm trying to build the 3rd party libraries needed to build FreeSurfer, and it seems that `numpy` and `nibabel` are imported from within the `build_packages.py` script.  I'd like to not have these modules installed in my environment before installing FS itself. Alternatively, this could be mentioned within the wiki page.

ltirrell:
This PR fixes #553  (although Andrew's commit 66d7075 also fixes this in a different way), adding a number of of changes to the `python/freesurfer` and `build_packages.py` code to allow packages to be built without importing `nibabel` or `numpy`, while still keeping nicer terminal logging. 

Also, within the build script, the `Package` objects aren't created until they are about to be built, so that the package tarballs/scripts don't need to exist if they aren't being built.  This is useful for the case of a minimal build, so that the tarballs for e.g. GUI-based packages aren't needed. 

Some stylistic (mostly [PEP8](https://www.python.org/dev/peps/pep-0008/)-based) changes are also introduced:
- using 4 spaces for indents
- alphabetizing imports, splitting them into sections for stdlib, 3rd-party, and local (`python/freesurfer`) imports
- using `python3` instead of `python`
- avoiding compound statements (multiple statements on the same line), so this:
  ``` python
  if os.path.exists(src_dir): shutil.rmtree(src_dir)
  ```
  becomes
  ``` python
  if os.path.exists(src_dir):
      shutil.rmtree(src_dir)
  ```
Additionally, `mri_seg_overlap` is added to the `.gitignore`

Let me know if you have any suggestions, or would prefer to keep the previous code styling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -229,6 +229,7 @@ mri_robust_register/mri_robust_register
 mri_robust_register/mri_robust_template
 mri_sbbr/mri_sbbr
 mri_seg_diff/mri_seg_diff
+mri_seg_overlap/mri_seg_overlap
 mri_segcentroids/mri_segcentroids
 mri_seghead/mri_seghead
 mri_segment/mri_segment
--- a/packages/build_packages.py
+++ b/packages/build_packages.py
@@ -1,10 +1,10 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # This is a utility script to help external developers easily build
 # freesurfer dependencies. The packages, defined below in the 'pkgs' list,
 # are built with the scripts and tarballs stored in the packages/source dir
 # and will get installed to the destination directory specified on the command-line.
-# 
+#
 # usage:
 #    build_packages.py /path/to/install/destination
 #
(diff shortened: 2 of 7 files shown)


buildqa:
I'll defer to Andrew on this one as he is the original author.

buildqa:
Closing this since 553 is closed above

ltirrell:
Is it worth doing a separate PR with the PEP-related style changes, and not requiring tarballs to exist unless you are building them (so you don't need GUI-related packages if you're only building minimal FS)?

ahoopes:
Sorry Lee, I lost track of this one. It probably makes the most sense to open a new PR at this point since most of the changes would have to be reverted, but I can reopen this if you want keep committing on top of your `enh-reqs` branch.

PR #561, opened 2018-10-14
Bf/nf: Downsampling of GCAMs

mu40:
Added flag for downsampling GCAMs by factor of 2 to mri_concatenate_gcam, and fixed bugs in utils/gcamorph.c:GCAMdownsample2():

(1) Data were not written to destination but to source GCAM.
(2) Coordinates of allocated destination GCAM were assumed, wrongly, to be initialised to zero.
(3) Interpolation asymmetry/shift was introduced by setting xd(n) to mean of xs(2n) and xs(2n+1).

When downsampling GCAMs, averaging neighbouring nodes is not necessary: during application, e.g. using GCAMmorphToAtlas(), a weighted mean of nodes is computed anyway, e.g. in GCAMsampleMorph(). Averaging twice makes the applied downsampled warp more different from the original warp.

Updated test for mri_concatenate_gcam; new complete test data can be found here: 
http://gate.nmr.mgh.harvard.edu/filedrop2/?p=agzrygshpxe

### DIFF
--- a/mri_concatenate_gcam/mri_concatenate_gcam.cpp
+++ b/mri_concatenate_gcam/mri_concatenate_gcam.cpp
@@ -35,6 +35,7 @@ struct Parameters
   std::vector<std::string> fileList;
   bool reduce = false;
   bool invert = false;
+  bool downsample = false;
 };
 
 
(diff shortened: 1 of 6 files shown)


ahoopes:
Just updated testdata 594d008bc90158e45e98dc0785f51b3d93f078ca

PR #562, opened 2018-10-16
Nf: mri_fuse_segmentations with GCAMs

mu40:
Added GCAMs support to mri_fuse_segmentations, with flexibility for specifying transform file names on the command line (as opposed to hard-coded paths).

Changed interface, ensuring behaviour for LTAs/identity transforms stays identical to that of the previous version.

Updated test since previous version compared empty images. Data: 
http://gate.nmr.mgh.harvard.edu/filedrop2/?p=6scafo594l0


buildqa:
Merged this in since checks passed and no conflicts.

PR #563, opened 2018-10-18
change pybind_python_version to 3.6

ltirrell:
Could the `PYBIND11_PYTHON_VERSION` be set to `3.6` by default? If `DISTRIBUTE_FS_PYTHON` is used (which is default), Python version 3.6 is installed, so should these two match?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -306,7 +306,7 @@ if(DISTRIBUTE_FSPYTHON)
 endif()
 
 # initialize pybind for python wrapping
-set(PYBIND11_PYTHON_VERSION 3.5)  # minimum version required
+set(PYBIND11_PYTHON_VERSION 3.6)  # minimum version required
 add_subdirectory(packages/pybind11)
 
 if(NOT DISTRIBUTE_FSPYTHON)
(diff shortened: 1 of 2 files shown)


ahoopes:
`PYBIND11_PYTHON_VERSION` is just the minimal version allowed, not the default. I imagine most people have python 3.6 installed at this point, but I think those that still have 3.5 should be able to configure freesurfer without errors. Ubuntu 16 is still commonly used, and that ships with 3.5

ltirrell:
Do you know what would happen if both 3.5 and 3.6 are installed on the system? For example, if 3.6 is set to be my `python3` (so all my pip installed packages are 3.6), but I also have the Ubuntu system python 3.5, which version would pybind use if `PYBIND11_PYTHON_VERSION` is set to 3.5? Or does it matter?

ahoopes:
I'm pretty sure it will use the first python >= 3.6 it finds in your path, so if your `python3` points to your 3.6 install, it will use that. Although if your `python` points to the system 3.5 (and not python2), then it probably depends on your path order

buildqa:
Closing this one; you can order your path to find the version of python that's needed.

PR #564, opened 2018-10-19
WIP: make code py2/3 compatible, use /usr/bin/env python

ltirrell:
**_WIP: Please don't merge yet_**
Uses [`futurize`](https://python-future.org/automatic_conversion.html) (with manual intervention) to convert (known) python in the packages/scripts dir to Python2/3 compatible code.

Note that `futurize` stage 2 can add a 3rd party dependency to the modules `builtins` or `past` from the package `future`. After inspecting all the automated changes, it seems that any code relying on this was unnecessary, so the imports were removed. These fell into the following categories:
- getting integer division (default for Python 2) in Python3, which didn't seem to be the intent of division anywhere in these changed files
- getting the `range` generator (default in Python 3), as opposed to the the list version of `range` in Python 2. again, doesn't seem necessary for these cases
- importing `from builtins import str`. It seems like nothing fancy was done with strings in these files, so that Py2/Py3 strings should be interchangable

### DIFF
--- a/packages/build_packages.py
+++ b/packages/build_packages.py
@@ -17,6 +17,7 @@
 # if this script is rerun, the package is only rebuilt when the source code has been
 # modified/updated.
 
+from __future__ import print_function
 import os
 import sys
 import platform
--- a/scripts/LongQdecTable.py
+++ b/scripts/LongQdecTable.py
@@ -1,6 +1,7 @@
 # Original author - Martin Reuter
 # $Id: LongQdecTable.py,v 1.5 2012/05/30 22:50:21 mreuter Exp $
 from __future__ import print_function
+from __future__ import division
 import os
 import logging
 import sys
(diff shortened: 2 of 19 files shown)


ltirrell:
It seems this is more difficult than I thought at first glance ...

There are more files I missed in the first pass. I pushed additional `futurize` stage 1 fixes, but stage 2 fixes are still WIP. 

Of note, the `StableDict` class defined in `datstruct_utils.py` and used in a number of other files modifies the base Python dict `dict`, and this approach is known not to work in Python 3. I haven't looked at the details on how difficult it would be to write a Py2/3 compatible version.
https://github.com/freesurfer/freesurfer/blob/5a3c4022fe9e766bed1cb0d71cb1070d6a7a7d0d/scripts/datastruct_utils.py#L144

ahoopes:
@ltirrell closing this for now. Feel free to reopen further down the line

PR #568, opened 2018-10-31
Nf: GCAM support for mri_mask

mu40:
Added GCAM support to mri_mask. To facilitate this, replaced '-invert' flag with automated transform inversion (if permitted by source and destination geometries). Flag was not used anywhere in the code base. Test data can be found here: http://gate.nmr.mgh.harvard.edu/filedrop2/?p=bmkp3e20qxg


mu40:
Updated test data. Please use this link: http://gate.nmr.mgh.harvard.edu/filedrop2/?p=aa33dvd5izr

PR #572, opened 2018-11-05
Nf: add samseg lesion feature

ste93ste:


buildqa:
Still waiting on some tests per Andrew before merging in.

PR #581, opened 2018-11-13
Bf: change to larger GCAM target image if downsampled

mu40:
Updated testdata: http://gate.nmr.mgh.harvard.edu/filedrop2/?p=2vsy1hv7wic


buildqa:
Added in and closing

PR #582, opened 2018-11-13
Bf: apply morph downsampled from odd number of slices

mu40:
The dimensions of a GCAM with spacing 1 equal the dimensions of its target/atlas. When the GCAM is downsampled to spacing 2, its size becomes floor(dimensions/2). If the target has an odd number of slices, now, the product of size and spacing no longer equals the target size.

To avoid misery, the size of the morphed image should therefore be set to the atlas size in GCAMmorphToAtlas().

Updated testdata for mri_convert: http://gate.nmr.mgh.harvard.edu/filedrop2/?p=2gwrbqbgv81


buildqa:
Adding in and closing.

PR #602, opened 2018-12-17
Bf more managed xyz and dist 1

emtpgeek:
Change a lot more modifies of xyz to use MRISsetXYZ

        Free the VERTEX::dist in many more places where the dist is invalid, to verify it is being recomputed with the new xyz values

        Use MRISexportXYZ rather than VERTEX::tdxyz in some cases to reduce usage of the tdxyz and get better memory traffic performance

        Add base.h copeWithLogicProblem that allows reporting and optional fixing of possible logic problems in the code

        Move origxyz after xyz in the VERTEX because xyz is the input and origxyz is the output

        Start keeping track of the dist_orig_nsize because changing dist_nsize did not always recompute the corresponding dist_orig values

        Replace calls to the badly misnamed MRIScomputeGeometricProperties with the existing MRIS_facesAtVertices_reorder, because that is what it does

        Passed __FILE__ and __LINE__ into MRISsetXYZ to allow for better warning messages

        Added and used MRISblendXYZandTXYZ   (mris, ...) rather than having multiple copies of the operation   (xyz = a * xyz + b * txyz)

        Added and used MRISscaleThenTranslate(mris, ...) rather than having multiple copies of the operation   (x = ax * x + bx), (y = ay * y + by), ...

        Use mrisComputeSurfaceDimensions rather than having multiple copies of the operation

        Logic errors being flagged
            mris_sphere/mris_sphere.c main()
                is not copying the xyz into the origxyz
                hence is later using zeroes rather than the correct values
            utils/mrisurf_integrate.c MRISinflateBrain()
                is not copying the xyz into the origxyz
                hence is later using zeroes rather than the correct values

        Add a primitive restart capability to recon-all
            so that the initial expensive mri section can be skipped when working on mris code

        Share the code for mrisComputeVertexDistance and mrisComputeOriginalVertexDistance
            so that they use the exact same definition of distance - the OriginalVertex version had not been updated when the Vertex version was sped up

        Speed up the deallocation and reallocation dist and dist_orig so it is cheap to have the dist and dist_orig fields NULL when the values are invalid

        Added mrisComputeOriginalVertexDistancesIfNecessary which will compute the dist_orig if they are not already computed
            This is useful for correctly logic errors where some paths had not computed the dist_orig and other paths might have

        Enhanced MRIS_facesAtVertices_reorder to fill in the topology correctly

        Renamed  MRIScomputeNormals  to  MRIScomputeNormalsPerturbingXYZ  because the former does not reveal the idea that the xyz might be changed by this call

### DIFF
--- a/dngtester/dngtester.c
+++ b/dngtester/dngtester.c
@@ -133,17 +133,22 @@ int main(int argc, char **argv)
   for(s=0; s < surf->nvertices; s++){
     v = &(surf->vertices[s]);
     for(c=0; c<3; c++){
-      if(c==0) v->x += delta;
-      if(c==1) v->y += delta;
-      if(c==2) v->z += delta;
+      double dx = (c==0) ? delta : 0.0;
+      double dy = (c==1) ? delta : 0.0;
+      double dz = (c==2) ? delta : 0.0;
+      MRISsetXYZ(surf,s,
+        v->x + dx,
+        v->y + dy,
+        v->z + dz);
       MRISfaceNormalGrad(surf, 0);
       MRISedgeGradDot(surf);
       //d1 = MRISbbrCost(bbrpar, NULL);
       d1 = MRISedgeCost(surf, NULL);
       gnum->rptr[1][c+1] = (d1-d)/delta;
-      if(c==0) v->x -= delta;
-      if(c==1) v->y -= delta;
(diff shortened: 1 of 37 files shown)


emtpgeek:
recon-all and test get the same results as before

PR #606, opened 2018-12-27
Build when tcl tk is installed on the system but is wrong

emtpgeek:
I have an Ubuntu 16.0 system with a tcltk8.6 installed.   When I try to build the dev stream packages directory it does not build.  Fundamentally it finds the wrong tcltk, and 8.6 is not compatible with the obsolete features used in vtk.

This is a patch that fixes it - but it is not quite the right patch, and (obviously) I have not tested it on other systems.

It changes build_packages.py to build tcltk before vtk, because vtk needs it

It adds a few lines that tell cmake where to look for the tcltk

It adds a few lines that tell make to include the tcl tk from the built packages directory

It puts a replacement CMakeLists.txt into the unzipped kwwidgets directory as part of this

It also runs cmake twice in kwwidgets before doing the make.   Weirdly cmake creates files, and some of those files need to be involved in process of correctly finding and using tcltk - I have not worked out precisely why

Andrew?  Other?  needs to look at this, and tweak it into a good fix.  But at least now I can build all the third party packages on my system with tcl tk 8.6 publicly installed


ahoopes:
Closing as tcltkblt and kwwidgets are no longer needed

PR #607, opened 2018-12-27
Bf freeview wouldnt link

emtpgeek:
Before this, attempts to build freeview on my Ubuntu 16 system failed with unresolved linker symbol messages

This fixes that by fixing the order of the libraries, and by including Xt in the link

freeview now links, but does not yet work - the Preferences window appears but is not populated

I don't know who the freeview developer is - obviously they need to look at this pull request before anything is done with it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,10 @@ if(BUILD_GUIS)
   # only used by mris_decimate_gui
   find_package(WXWidgets)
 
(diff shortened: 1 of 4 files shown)


emtpgeek:
This also fixes a few of the other utilities that would not link against the built packages

ahoopes:
Closing as this was resolved in https://github.com/freesurfer/freesurfer/pull/612

PR #609, opened 2018-12-31
Replace all instances of Thalamus-Proper with Thalamus

ltirrell:
Follows up on #601, this replaces any use of `Thalamus-Proper` within the repo with `Thalamus`, using:
```bash
sed -i s'/Thalamus-Proper/Thalamus/'g $(find . -type f -not -path "./.git*")
```
from the top-level directory.

@ahoopes it doesn't seem that this will cause issues, as it seems that most cases refer to the segIDs of `Thalamus` (10, 49) in the changed code. However, not 100% sure if this renaming will cause issues with Brainstem/HippoSF/samseg code that I'm unaware of


ahoopes:
thanks @ltirrell - this might impact samseg a bit, so I'm running some tests before merging

ahoopes:
ca0669fea7963d500b1e3b63f6a609c092a25314 adds to this by removing `-Proper` references in the average subjects (everything but the recon logs)

PR #610, opened 2019-01-05
Fix Test Data tarballs

freesurfer-rge:
Add CMake commands to copy the various tarballs of test data to the proper locations
This is to allow for out-of-source builds and test runs


ahoopes:
Closing now that out-of-source build testing is fully supported

PR #621, opened 2019-01-29
This is obviously a typo

emtpgeek:
and yet it makes no difference at all to the tests and the recon-all output!

Lets fix it now before it causes weird problems in the future


dngreve:
@emtpgeek Bevin, are you getting our emails? We've sent several but have not heard back.
thanks
doug
@ahoopes @dngreve

emtpgeek:
Maybe not - the last email I had from Doug was 1/16 about an mris_decimate problem that has been (I believe) since fixed
________________________________

PR #626, opened 2019-02-05
Improve behaviour of recon-all buckner 144

emtpgeek:
Same results on recon-all bert

When adding more faces to a defect fix, don't add more than two faces to an edge

Adds consistency checking code as well, but it is disabled


ahoopes:
Temporarily disabling this fix since it produces failures when fixing longitudinal surfaces (574f28f05436940cac81d3a95b9ee2d4230e24cd). The exact source of this issue is still unknown.

PR #635, opened 2019-03-06
Reenable this fix, fix the crash it caused, fix mris_diff to be able to compare

emtpgeek:
The original problem was patching creating more than two faces attached to an edge

Once this problem was fixed,  it became possible for a vertex to have no faces attached to it, which caused an assertion.  Easily fixed.

But then the new version of the patching code is creating a different surface - less edges,vertices,faces - than the previous version.  So the mris_diff was failing.

This code therefore also fixes mris_diff to be able to compare surfaces with completely different vertex and face numbering, and a few vertices missing.  It relies on most of the vertices being at the same unique xyz as the old surface.

It does so by matching the same-location vertices, then matching the faces, then allowing a small number of differences.

Gets same results on recon-all bert

Passes the same tests as dev


ahoopes:
In general, if two surfaces differ in the number of vertices or faces, we'd want mris_diff to fail no matter how similar the surfaces are. This new feature you've added is quite helpful, but I think it would be best to implement it as an option that must be specified on the command line. If the fixed mris_fix_topology is failing the test comparison, then that's just a matter of regenerating the test data.

emtpgeek:
Ok - I'll put a switch for it, defaulted to be off

ahoopes:
great thanks! I'll regenerate the test reference

PR #638, opened 2019-03-12
Bf freesurfer replacement for creation time string

emtpgeek:
I am seeing bad text in the currentDate field in the triangle files under obscure circumstances. This was causing crashes when the triangle file was read.

The code has a function call returning a std::string and the caller then taking its c_str().  The function result is then destroyed, but a pointer to its c_str has been kept and is later used.

This fix eliminates the problem.


ahoopes:
thanks!

PR #644, opened 2019-04-13
Should store these even if not previously stored

emtpgeek:
Does not change test results or recon-all of bert

Found this while investigating the mris_register diff, but affect that either

These should have been saved regardless of whether the dist_orig had already been computed


emtpgeek:
Will merge this into the next pull request

PR #651, opened 2019-04-23
Nf step towards faster deform and integrate merge dev

emtpgeek:
Identical results in recon-all bert and testing

This fixes a build problem introduced into the dev stream where uses of isnan and isinf were not compiling on my system

This begins to separate the surface projection and SSE term calculation out into functions that can be (future) turned into templates that can support multiple MRIS representations


emtpgeek:
Updating to latest dev

PR #659, opened 2019-05-05
isnan and a few others are stopping a build on ubuntu-16, this is the trivial fix

emtpgeek:
These aren't present in the header files visible in a few places, so this defines them without needing a "using namespace std"


ahoopes:
Closing as this has been included in https://github.com/freesurfer/freesurfer/pull/660

PR #661, opened 2019-05-06
nf: neighborhood euler number

ahoopes:


ahoopes:
Closing for now as it's not urgent and should really be refactored in a cleaner way

PR #681, opened 2019-06-22
Prepare to use Surface et al in SSE calculations

emtpgeek:
This doesn't introduce Surface et al into sse, but prepares to.  No changes to recon-all or tests.

Moves all the SSE and dx-dy-dz calculations out of mrisurf_deform.cpp into their own pair of files, and puts the SSE terms into a class that (next step) will be made a template that supports multiple representations.


ahoopes:
Thanks @emtpgeek. For debugging purposes (mostly in freeview) Doug and I were planning on breaking apart some of these per-surface SSE functions to easily compute terms on a per-vertex basis. For example, `mrisComputeNormalSpringTerm` would loop through the vertices but call some other function to actually compute the term for each vertex. Any thoughts on a clean way to implement this type of thing in the new surface framework without sacrificing speed?

emtpgeek:
Any thoughts on a clean way to implement this type of thing in the new surface framework without sacrificing speed?

Perhaps the following will show the composition of various ideas to allow us to be efficient and easily read and easily maintained...


#include "pch.h"
#include <iostream>
#include <algorithm>

// traditional algorithm but using C data accesses,
// making it hard to change data representations
//
namespace Traditional {
struct VERTEX {
 float p1, p2;
};
struct MRIS {
 int nVertices;
 VERTEX* vertices;
};
double term1(MRIS* mris) {
 double sum = 0.0;
 for (int i = 0; i < mris->nVertices; i++) sum += mris->vertices[i].p1;
 return sum;
}
double term2(MRIS* mris) {
 double sum = 0.0;
 for (int i = 0; i < mris->nVertices; i++) sum += mris->vertices[i].p2;
 return sum;
}
double computeSSE(MRIS* mris) { return term1(mris) + term2(mris); }
}

// traditional algorithm but using C++ member functions etc. to access the data
// A good compiler will produce equally efficient code to the above,
// but it is now possible to choose different data formats without changing the algorithm
//
namespace Traditional_algorithm_SurfaceUsingMRIS {
struct Vertex {
 Traditional::MRIS* mris;
 int i;
 Vertex() : mris(nullptr), i(0) {}
 Vertex(Traditional::MRIS* mris, int i) : mris(mris), i(i) {}
 float p1() { return mris->vertices[i].p1; }  float p2() { return mris->vertices[i].p2; }
};
struct Surface {
 Traditional::MRIS* mris;
 Surface(Traditional::MRIS* mris) : mris(mris) {}
 int nVertices() { return mris->nVertices; }
 Vertex vertices(int i) { return Vertex(mris,i); }
};
double term1(Surface mris) {
 double sum = 0.0;
 for (int i = 0; i < mris.nVertices(); i++) sum += mris.vertices(i).p1();
 return sum;
}
double term2(Surface mris) {
 double sum = 0.0;
 for (int i = 0; i < mris.nVertices(); i++) sum += mris.vertices(i).p2();
 return sum;
}
double computeSSE(Surface mris) { return term1(mris) + term2(mris); }
}

// traditional algorithm but using templates to access the C++ member functions etc.
// A good compiler will produce equally efficient code to the above,
// but it is now can use several representations without changing the algorithm
//
namespace Traditional_algorithm_Templated_Surface {
template <class Vertex, class Surface>
struct Terms {
 double term1(Surface mris) {
  double sum = 0.0;
  for (int i = 0; i < mris.nVertices(); i++) sum += mris.vertices(i).p1();
  return sum;
 }
 double term2(Surface mris) {
  double sum = 0.0;
  for (int i = 0; i < mris.nVertices(); i++) sum += mris.vertices(i).p2();
  return sum;
 }
};

struct VectorBasedMRIS {
 float* p1;
 float* p2;
 size_t size;
};
struct Vertex {
 VectorBasedMRIS* mris;
 int i;
 Vertex(VectorBasedMRIS* mris, int i) : mris(mris), i(i) {}
 float p1() { return mris->p1[i]; }  float p2() { return mris->p2[2]; }
};
struct Surface {
 VectorBasedMRIS* mris;
 Surface(VectorBasedMRIS* mris) : mris(mris) {}
 int nVertices() { return mris->size; }
 Vertex vertices(int i) { return Vertex(mris, i); }
};

double computeSSE(VectorBasedMRIS* mris) {
 Terms<
  Vertex,
  Surface> terms;
 Surface surface(mris);
 return terms.term1(surface) + terms.term2(surface);
}

double computeSSE(Traditional::MRIS* mris) {
 Terms<
  Traditional_algorithm_SurfaceUsingMRIS::Vertex,
  Traditional_algorithm_SurfaceUsingMRIS::Surface> terms;
 Traditional_algorithm_SurfaceUsingMRIS::Surface surface(mris);
 return terms.term1(surface) + terms.term2(surface);
}
}


// This can now be extended to allow the evaluation of the terms for just some vertices
//
// The approach here does involve running the loops separately, which is okay when
// the loop control flow is tiny compared to the loop body, which I think is going to
// be the case for most of our terms.
//
namespace Traditional_algorithm_Templated_Surface_Improved {
template <class Vertex, class Surface>
struct Terms {

 // By supplying a begin end we can use the same code to do 1, a chunk, or all
 //
 double term1(Surface mris, int begin, int end) {
  double sum = 0.0;
  for (int i = begin; i < end; i++) sum += mris.vertices(i).p1();
  return sum;
 }
 double term2(Surface mris, int begin, int end) {
  double sum = 0.0;
  for (int i = begin; i < end; i++) sum += mris.vertices(i).p2();
  return sum;
 }

 // we can use chunking to do several terms while efficiently using the cache
 //
 double term1_plus_term2(Surface surface, int begin, int end) {
  double sum = 0.0;
  for (int lo = begin; lo < end; lo += 128) {
   auto hi = std::min(lo + 128, end);
   sum += terms.term1(surface, lo, hi) + terms.term2(surface, lo, hi);
  }
 }
};

struct VectorBasedMRIS {
 float* p1;
 float* p2;
 size_t size;
};
struct Vertex {
 VectorBasedMRIS* mris;
 int i;
 Vertex(VectorBasedMRIS* mris, int i) : mris(mris), i(i) {}
 float p1() { return mris->p1[i]; }  float p2() { return mris->p2[2]; }
};
struct Surface {
 VectorBasedMRIS* mris;
 Surface(VectorBasedMRIS* mris) : mris(mris) {}
 int nVertices() { return mris->size; }
 Vertex vertices(int i) { return Vertex(mris, i); }
};

double computeSSE(VectorBasedMRIS* mris) {
 Terms<
  Vertex,
  Surface> terms;
 Surface surface(mris);

 return terms.term1_plus_term2(surface,0,surface.nVertices());
}

double computeSSE(Traditional::MRIS* mris) {
 Terms<
  Traditional_algorithm_SurfaceUsingMRIS::Vertex,
  Traditional_algorithm_SurfaceUsingMRIS::Surface> terms;
 Traditional_algorithm_SurfaceUsingMRIS::Surface surface(mris);
 return terms.term1_plus_term2(surface, 0, surface.nVertices());
}
}


// This can now be extended to allow the evaluation of the terms for just some vertices
//
// for critical cases where the per-vertex calculation is tiny,
// we can merge some of the term calculations.
//
namespace Traditional_algorithm_Templated_Surface_Improved {
template <class Vertex, class Surface>
struct Terms {

 // By supplying a begin end we can use the same code to do 1, a chunk, or all
 // This code relies on the branch prediction hardware to eliminate
 // the cost of the if (term#) test in the loop body.  If there are more than
 // a few terms, then they can be grouped and an if statement first check for the
 // whole group
 //
 void terms_1_and_2(Surface mris, int begin, int end, double* term1, double* term2) {
  double sum1 = 0.0, sum2 = 0.0;
  for (int i = begin; i < end; i++) {
   if (term1) sum1 += mris.vertices(i).p1();
   if (term2) sum2 += mris.vertices(i).p2();
  }
  if (term1) *term1 = sum1;
  if (term2) *term2 = sum2;
 }
 double term1(Surface mris, int begin, int end) {
  double sum;
  terms_1_and_2(mris, begin, end, &sum, nullptr);
  return sum;
 }
 double term2(Surface mris, int begin, int end) {
  double sum;
  terms_1_and_2(mris, begin, end, nullptr, &sum);
  return sum;
 }

 double term1_plus_term2(Surface surface, int begin, int end) {
  double sum1 = 0.0, sum2 = 0.0;
  for (int lo = begin; lo < end; lo += 128) {
   auto hi = std::min(lo + 128, end);
   terms_1_and_2(surface, lo, hi, &sum1, &sum2);
  }
  return sum1 + sum2;
 }
};

struct VectorBasedMRIS {
 float* p1;
 float* p2;
 size_t size;
};
struct Vertex {
 VectorBasedMRIS* mris;
 int i;
 Vertex(VectorBasedMRIS* mris, int i) : mris(mris), i(i) {}
 float p1() { return mris->p1[i]; }  float p2() { return mris->p2[2]; }
};
struct Surface {
 VectorBasedMRIS* mris;
 Surface(VectorBasedMRIS* mris) : mris(mris) {}
 int nVertices() { return mris->size; }
 Vertex vertices(int i) { return Vertex(mris, i); }
};

double computeSSE(VectorBasedMRIS* mris) {
 Terms<
  Vertex,
  Surface> terms;
 Surface surface(mris);

 return terms.term1_plus_term2(surface, 0, surface.nVertices());
}

double computeSSE(Traditional::MRIS* mris) {
 Terms<
  Traditional_algorithm_SurfaceUsingMRIS::Vertex,
  Traditional_algorithm_SurfaceUsingMRIS::Surface> terms;
 Traditional_algorithm_SurfaceUsingMRIS::Surface surface(mris);
 return terms.term1_plus_term2(surface, 0, surface.nVertices());
}
}





int main()
{
    std::cout << "Hello World!\n";
}


________________________________

ahoopes:
@emtpgeek so in this case `surface.vertices(i).p1()` is returning an already computed value? How could these terms actually be calculated - via member functions of a vertex as well? Currently they're computed all at once by a single function, but ideally we'd want to define functions that compute terms on a per-vertex basis.

emtpgeek:
I submitted a pull request today that starts down the route I think you are interested in

I have only implemented it for one of the terms, but the mechanism is in place for more than one,  which will be the next step

It adds a trailing defaulted arg to, for instance, mrisComputeRepulsiveEnergyRatio(..., int selector=-1), which causes the term to be summed over all vertices if -1 is passed, but computes the term for just one vertex if its vno (or fno, for terms that iterate over faces) is passed.

I can't think of a case where we want some partial set of vertices to be summed over, but that would be easy to add also

/Bevin
________________________________

ahoopes:
@emtpgeek distance term functions, like `mrisComputeNormalSpringTerm()`, are different than energy terms since they set/append the vertex dxyz during surface placement instead of just returning a total sum. These are the functions that we'd like to break apart in order to compute per-vertex values - i.e. one function will compute the distance term for a given vertex while another will actually set this `dx/dy/dz` for each vertex.

Say there is a function (maybe a vertex-class member-function) like `normalSpringTerm()` that returns dxyz values for a given vertex. We could have a  another term-specific function, equivalent to the current `mrisComputeNormalSpringTerm()`, that loops through each vertex and sets the `v->dxyz` to the result of `v->normalSpringTerm()`, or something like that. Or we could also have a more generic (perhaps templated) function that is capable of doing for this for all distance terms.

Do you think that mimicking the new SSE framework is the way to go in this case?

emtpgeek:
We have two fundamentally different operations

  1.  calculate a dxyz for each vertex
  2.  use the dxyz for all vertices to mutate the surface

As you say, it would be very nice to separate these out, so (1) the dxyz can be calculated for a subset of the vertices without modifying the global structures, and (2) the computed dxyz can be used later to mutate the surface

This decision will be easier to implement if these operations are changed to using the new SSE framework

I'll do that in the next pull request

/Bevin
________________________________

emtpgeek:
The change you made to the utils/mrisurf_compute_dxyz.cpp is right on target

I am looking at the various terms and notice that they fall into two categories
1:  loops with a trivial amount of work per vertex
2:  loops with a complicated algorithm applied to each vertex

There may be a case for a framework that lets us compute several trivial terms during the same loop, and the inline functions will enable that should it turn out to be the case

/Bevin
________________________________

PR #689, opened 2019-07-13
Bf increase max neighbors

emtpgeek:


emtpgeek:
No changes to results or recon-all

Raises MAX_NEIGHBORS to a number bigger than that seen in the recently reported problem with freeview

Encapsulates the changing of VERTEX_TOPOLOGY vnum into a single function to make it easy to find where it is being set to very high values, to see that it was in the reading from the input file and not from some later incorrect analysis or corruption

Adds code that traced the setting of the very high number of neighbors so it could be seen to be legit

PR #710, opened 2019-09-17
Update suggestion for load_nifti.m

schoffelen:
I propose that the nifti-reader for matlab tries to preserve the input datatype, if possible. This can be achieved with the suggested PR. Feel free to discard, if it has unwanted ramifications.


schoffelen:
This is getting too messy, apologies for that: I will close and file a new, clean PR

PR #711, opened 2019-09-18
proposed changes to reading in nifti and mgh files into matlab

schoffelen:
I propose to preserve the datatype of the data in the input volumes, rather than converting all to double precision floats. This saves a lot of memory potentially when reading 4-dimensional files.

### DIFF
--- a/matlab/load_mgh.m
+++ b/matlab/load_mgh.m
@@ -195,13 +195,13 @@ end
 if(slices(1) <= 0 & frames(1) <= 0)
   switch type
    case MRI_FLOAT,
-    vol = fread(fid, nv, 'float32') ; 
+    vol = fread(fid, nv, '*float32') ; 
    case MRI_UCHAR,
-    vol = fread(fid, nv, 'uchar') ; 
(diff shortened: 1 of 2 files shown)


ahoopes:
@schoffelen thanks! This is useful, but often times matlab functions expect input data with type double. You can just do `double(data)` in this case, but it defeats some of the ease of using matlab and could interfere with others' code. I've added a switch in commit fa286561fde0cb072657806f9401012dfc6bb033 to enable your changes. To have matlab preserve volume datatype, you can set the following environment variable:

```bash
export FS_PRESERVE_MATLAB_VOLTYPE=1
```

Otherwise, the previous behavior will be used by default.

schoffelen:
Thanks for accommodating this. Good point indeed about backward compatibility. In our software (Fieldtrip) we try to check for and enforce double/single precision before embarking on the relevant computations, but that cannot be said with certainty for all applications. I'll make the required adjustments on our end to optionally switch on/off the flag then (most users of Fieldtrip are not aware they are using freesurfer under the hood, so they will not have set up their environment appropriately).

PR #728, opened 2019-11-07
nf: samseg lesion add gaussian lesion tied to gaussian wm 

ste93ste:
bf: fix normalization term in variance updates


ahoopes:
This produces different results in the cross stream. Is this expected / presumably related to the fixed normalization term?

ste93ste:
Yes, it's due to the fixed normalization term. The results might be different for few voxels (especially for small structures).

ahoopes:
Sounds good - the differences don't appear to be significant

PR #748, opened 2020-02-06
testing travis...

ahoopes:
Travis is building branches but not sending status back to github. I've tried uninstalling and reinstalling... let's see if that works


ahoopes:
guess that worked!

PR #753, opened 2020-02-09
rf: rename files correctly inside comments; makes for better tooling and searching

aboualiaa:
This was done by a script, so a few 'dummy.cpp's have slipped through


ahoopes:
thanks for this, but closing because we're actually in the process of replacing all of this header information from the source files

PR #762, opened 2020-02-25
rf: remove dead code from MRIchangeType

aboualiaa:
The checks before MRIsetVoxVal are redundant, since they are done inside it.
Remove dead code that has been replaced by MRIgetVoxVal


ahoopes:
thanks!

PR #771, opened 2020-04-26
Remove -Wno-absolute-value and -Wno-write-strings

freesurfer-rge:
Remove two suppressions for compiler warnings - `-Wno-absolute-value` and `-Wno-write-strings`. The first was straightforward, the second less so.

To sort out `-Wno-write-strings`, it was necessary to change declarations such as `char* hemi = "lh";` into `const char* hemi = "lh";`. Where these were passed in to subroutines, the signatures had to be adjusted as well (or in a few cases, a `const_cast` used). In general, occurrences of `char *` should probably be replaced with `std::string`. There are places (and the `hemi` variable actually appears to be a good candidate) where an `enum class` (with suitable `ToString()`  and `Parse()` methods) would be more appropriate.


include/mrisutils.h
```diff
@@ -51,11 +51,11 @@ typedef struct {
   int nsubjects;
   MRIS *targsurfreg; // an actual surface to use as the target registration
   int  icoorder; // only use if targsurfreg not spec
-  char *targsubject; // only use if targsurfreg and icoorder not spec
-  char *hemi; // lh or rh
-  char *surfname; // eg, white
-  char *surfregname; // eg, sphere.reg
-  char *xform_name; // eg, talairach.xfm
+  const char *targsubject; // only use if targsurfreg and icoorder not spec
+  const char *hemi; // lh or rh
```
freesurfer-rge, line 55: Given that the values have to be `lh` or `rh` an `enum class` with `ToString()` and `Parse()` methods might be better.

mri_gca_ambiguous/mri_gca_ambiguous.cpp
```diff
@@ -61,7 +61,7 @@ static double MIN_FA3 = 1  ;
 static  double MAX_FA3 = 40 ;
 static  double FA_STEP = 1 ;
 static int append = 0 ;
-static char *fname = "amb.log" ;
+static const char *fname = "amb.log" ;
```
freesurfer-rge, line 64: I wasn't sure why sometimes declarations like this were `static` and sometimes not.

mri_probe_ima/mri_probe_ima.cpp
```diff
@@ -549,7 +550,7 @@ static void check_options(void) {
   return;
 }
 /*-------------------------------------------------------------*/
-static int stringmatch(char *s1, char *s2) {
+static int stringmatch(const char *s1, const char *s2) {
```
freesurfer-rge, line 553: This method appears to be duplicated in a lot of files (there are variants which also get duplicated). I didn't refactor them because, if `char *` is replaced by `std::string` then the equality operator can be used instead (and all of these removed).

mri_seghead/mri_seghead.cpp
```diff
@@ -47,9 +47,9 @@ static void print_help(void) ;
 static void print_version(void) ;
 static void argnerr(char *option, int n);
 static void dump_options(FILE *fp);
-static int  isflag(char *flag);
+static int  isflag(const char *flag);
```
freesurfer-rge, line 50: I saw this routine in a number of files as well. Statically declared and (apparently) copied between them.

mris_congeal/mris_congeal.cpp
```diff
@@ -486,7 +486,7 @@ get_option(int argc, char *argv[])
            "and using it for initial alignment\n", inflated_name) ;
     sprintf(fname, "%s.H", argv[2]) ;
     curvature_names[0] = (char *)calloc(strlen(fname)+1, sizeof(char)) ;
-    strcpy(curvature_names[0], fname) ;
+    strcpy(const_cast<char*>(curvature_names[0]), fname) ; // const_cast AND strcpy
```
freesurfer-rge, line 489: Note that `curvature_names` is an array of character arrays declared and initialized above. If `fname` is ever longer than the first entry in this array, I believe the `strcpy()` is going to clobber something random. In any case, I believe that it's going to clobber the original constant initialisation.

mris_register_to_volume/mris_register_to_label.cpp
```diff
@@ -125,7 +125,7 @@ static void argnerr(char *option, int n);
 static void dump_options(FILE *fp);
 static int  singledash(char *flag);
 #include "tags.h"
-static int istringnmatch(char *str1, char *str2, int n);
+static int istringnmatch(const char *str1, const char *str2, int n);
```
freesurfer-rge, line 128: I believe this pops up in quite a few files. Again, I didn't refactor because if `std::string` is used instead, then there's probably better alternatives.

freesurfer-rge:
If this is merged, please squash and use the PR description for the merge, rather than the commit comments.

ahoopes:
Thanks, this is a huge help! I'm merging now but will reenable the `-Wno-write-strings` flag globally for now since there are still invalid string literal conversions in `mris2rgb`, `stim_polar`, and `mri_probedicom`, which weren't checked by CI since they require OpenGL

PR #811, opened 2020-09-16
rf: hypothlamic_subunits updates (posteriors, volumes, threads)

BBillot:
just a couple more changes:
- posteriors are now optional in FS mode
- if several subjects are run in the FS mode, a file with the volumes of all subjects is saved in $SUBJECTS_DIR
- the number of used threads now defaults to 0


BBillot:
Sorry threads defaults to 1...

PR #888, opened 2021-10-21
dcm2niix, DICOMRead3 addition

yhuang43:
1. Added packages/dcm2niix, modified dcm2niix oct 6, 2021 version to provide interface for Freesurfer mri_convert
2. Added -dicomread3 option to mri_convert
    a. DICOMRead3() - use dcm2niix to convert DICOM to nii
    b. niiRead3() -  use output from DICOMRead3() to generate mgz

### DIFF
--- a/include/DICOMRead.h
+++ b/include/DICOMRead.h
@@ -27,6 +27,9 @@
 #include "dicom_objects.h"
 #include "condition.h"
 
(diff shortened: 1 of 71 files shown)


yhuang43:
I just realized that I checked in a wrong file. I need to do another pull request?

PR #918, opened 2022-01-24
mris_gradient_unwarp changes

yhuang43:
1. add option '--nthreads' to mris_gradient_unwarp, set omp_set_num_threads(nthreads)
2. separate GradUnwarp class into GradUnwarp and Siemens_B
3. create new Siemens_B object for each GradUnwarp::spharm_evaluate() call
4. array Alpha, Beta, &P start at index 1


yhuang43:
Doug would like to have array start at index 0

PR #940, opened 2022-02-25
Changes to the build to compile 2 versions of `libkvlGEMSCommon.a`

pwighton:
Samseg relies on `libkvlGEMSCommon.a`, which can be compiled with our without the flag `-DUSE_DYNAMIC_MESH`.

Enabling `-DUSE_DYNAMIC_MESH` is required to compile the following executables:
  - `kvlBuildAtlasMesh`
  - `kvlViewMeshCollectionWithGUI`
  - `kvlEstimateAtlasParameters`

Otherwise it is not needed, causes samseg to run up to 100% slower, and breaks the python bindings.

This PR changes the build so that two versions of the library is built: `kvlGEMSCommom` and `libkvlGEMSCommon_dynmesh`. The 3 executables above now link to `libkvlGEMSCommon_dynmesh` and everything else links against the faster `kvlGEMSCommom`.


pwighton:
Converting to draft because I just noticed the md5sum of `libkvlGEMSCommon.a` and `libkvlGEMSCommon_dynmesh.a` are identical

pwighton:
This has been tested and is ready to merge.

PR #1003, opened 2022-06-15
Additional argument for singularity command

Hoda1394:
I added "-e" to avoid errors when the user has freesurfer installed.
Also, the container couldn't use the gpu when "--gpu" flag was passed. So, "--nv" argument solved it.


ahoopes:
thank you @Hoda1394 !

PR #1013, opened 2022-08-01
Search for v13, then v12 of thalamic nuclei segmentation

ayendiki:


iPsych:
the latest subregion segmentation (beta) using
 segment_subregions
command provides 'ThalmicNuclei.FSvoxelSpace.mgz without version number.

PR #1023, opened 2022-09-07
removed non-decorator surfa printout

jnolan14:
Removed a 'SURFA NOTE' printout that was cluttering Bruce's terminal


avnishks:
We should've merged this earlier. 
Closing this now as Its stale, since Bruce pushed similar changes.

PR #1044, opened 2022-12-22
Fixed mri_synthsr: would crash because np.int is not defined anymore

jeiglesias:


avnishks:
The mri_synthsr_hyperfine script will need a similar change too [here](https://github.com/freesurfer/freesurfer/blob/dev/mri_synthsr/mri_synthsr_hyperfine#L870).

PR #1070, opened 2023-02-13
Update deprecated function on mri_3d_photo_recon

Jon-Williams:
Changed get_data() function to get_fdata() since the former is deprecated and throws an exception.


hvgazula:
@Jon-Williams Quick note (a recommendation): whenever you create a PR, please make sure to tag the right personnel (in this case @rpwang17 😄) so the developers know who it is intended for. Also, not everyone on the team may recognize us.

rpwang17 (approved):

hvgazula:
Thanks @rpwang17

PR #1127, opened 2023-08-08
ENH - matlab code; build in support for multiframe dicom files

schoffelen:
This is a long shot, and feel free to ignore, but this PR implements support for multiframe dicoms in the matlab reader.

Context: the Siemens Skyra scanner at the Donders Institute now natively creates multiframe dicoms since a recent software update. Our FieldTrip code (https://github.com/fieldtrip/fieldtrip) relies on load_dicom_fl.m for the reading of dicom anatomical volumes. Given the different formatting of the header (and data) this now fails for obvious reasons. 

I could easily implement a workaround that does not touch freesurfer functions, but on the off chance that you find this addition useful, I still submit it here, realizing that:
- the matlab code probably does not have the focus of attention of the developers
- core functionality (e.g. mri_convert) does not seem to support multiframe dicoms, so it could have been a conscious design choice not to support this particular file format (although I couldn't find anything about this when doing a shallow online search)
- I only tested the changed code on a few files that were collected at the Donders Institute, so I don't know how robust it is

### DIFF
--- a/matlab/load_dicom_fl.m
+++ b/matlab/load_dicom_fl.m
@@ -45,7 +45,7 @@ end
 nfiles = size(flist,1);
 
 tic
-fprintf('Loading dicom info foreach file \n');
+fprintf('Loading dicom info for each file \n');
 for n = 1:nfiles
   fname = deblank(flist(n,:));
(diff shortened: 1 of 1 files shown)


schoffelen:
Hello, is there anything related to this PR which prevents it from being picked up by the developers?

PR #1147, opened 2023-11-27
Make req.sh more generic - remove assumption that the source and build trees are one.

avnishks:
Make req.sh (python package management) work with separate source and build trees. 

FS-devs, let's discuss this issue here before merging.


avnishks:
Closing this PR as it's now redundant after commit [9b5e4ca](https://github.com/freesurfer/freesurfer/commit/9b5e4cafbfe8d9cb4b3b68a3aae369ea1b436306).

PR #1150, opened 2023-12-04
Update VolumeFilterThreshold.cpp

### LINKED ISSUE #1149: SetInput in has been renamed to SetInputData in vtkImageAlgorithm.h since VTK v.6.0.0
SetInput is still used in this file, which causes a compilation error.

https://github.com/freesurfer/freesurfer/blob/c385e43e3b250ec6c7fcdd6862f90c9dd0587bd4/freeview/VolumeFilterThreshold.cpp#L42

here is the commit from vtk:
https://github.com/Kitware/VTK/commit/4ea36770118ff350051d9d5eaf53c3a8de78a77d

the fix is a very simple rename, unless support for VTK v5 is necessary, but the packages which are officially distributed contain VTK v7, so I think it's safe to say that VTK v5 is not supported and should just be dropped from the code

aboualiaa:
Use SetInputData instead of SetInput since it has been renamed since VTK v6.0.0
Fixes #1149

### DIFF
--- a/freeview/VolumeFilterThreshold.cpp
+++ b/freeview/VolumeFilterThreshold.cpp
@@ -39,7 +39,7 @@ bool VolumeFilterThreshold::Execute()
 {
   TriggerFakeProgress(50);
   vtkSmartPointer<vtkImageThreshold> filter = vtkSmartPointer<vtkImageThreshold>::New();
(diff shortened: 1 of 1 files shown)


yhuang43:
@rpwang17 Can you review the proposed fix?

PR #1161, opened 2024-01-05
samseg lesion: expose do-not-use-shape-model (VAE) to user

ste93ste:
also small bf: set default threshold search string and value in samseg longitudinal lesion


yhuang43:
Hi Stefano, 

I have a question regarding the sampler parameter.

I see that ‘sampler=not args.do_not_use_shape_model’ is passed to both SamsegLesion and SamsegLongitudinalLesion constructors. But SamsegLongitudinalLesion doesn’t seem accepting the parameter. And SamsegLongitudinalLesion.constructSstModel() has the option hard coded to False. 

Should we give user the option to switch it on?

Yujing

ste93ste:
Hi Yujing,

Good catch! I copied the changes from the standalone version, but the version here did not have the sampler parameter. It should be fixed now.

The sampler=False in SamsegLongitudinalLesion.constructSstModel() is a model choice. It doesn't need to be exposed to the user.

Stefano

PR #1165, opened 2024-01-23
Add environment variable FS_LICENSE to Dockerfile

oikosohn:
Add environment variables(`FS_LICENSE`) to prevent errors in missing licenses


buildqa:
There is no longer a required or default location for the license file.

We don't provide license.txt or .license files as part of any freesurfer distribution. We expect end users to receive and download their own license.txt file after completing the online registration form. Then subsequently set FS_LICENSE to wherever they decide to download and/or move license.txt to. Many users opt for their $HOME directory or some location not under the freesurfer tree which does not require root privs to access.

PR #1175, opened 2024-02-13
bf: mri_synthmorph: fix init transform variable name typo

mccle:
In L542 of `mri_synthmorph/mri_synthmorph`, the variable `mov_to_new` is defined but never referenced again. In the full context of the code, I believe that this line is intended to redefine the `mov_to_net` variable to take the initial matrix transform into account and that this is simply a typo in the variable name. If I am correct, then this change should address bugs related to the application of initial transforms, but of course this should be verified.

In L643, I also corrected "mri_synthmoph" to be "mri_synthmorph", but this should not have any effect on code function.


mu40:
Thank you for identifying and addressing this bug! You are correct that the misspelled variable name interfered with the application of initial transforms, in that these were never incorporated into the deformation from the fixed to the moving image.

PR #1187, opened 2024-03-07
BF: Only copy the nested spheremorph folder

avnishks:


yhuang43:
There is a setup.py in the directory. Should we install the package using that like 'qatools' does?

avnishks:
We could! Spheremorph's installation script needs testing/works to make it compatible with all the combinations of different `OS` and `fspython` installation supported by FS.

This PR is essentially a short-term fix to unblock Rob. He is running into problems with linux installation tooling because, in it's current form, these two packages get copied over with similar named nested folders:
`python/packages/ERC_bayesian_segmentation/ERC_bayesian_segmentation`
`python/packages/spheremoprh/spheremorph`

yhuang43:
Ok. Just curious, do current scripts need to be updated to reflect the change?

avnishks:
Probably. I still another commit to add to this PR before I can merge. Will update any scripts if needed. Thanks!

jnolan14:
Let me know if you need any help looking into the ERC_bayesian files.

PR #1192, opened 2024-03-19
bf: Fix tensorflow macos arm compatability

htregidgo:
Fixes issues with some of the tensorflow based tools when running on an arm based macos machine (M1, M2, M3 chipset). The two main issues are:

	1. a check for np.float128 which isn't a supported datatype on this chipset
	2. an error in the shape of tensors provided to batch- normalisation layers

Issue 1 can be fixed by removing the check as it's not likely to be necessary in most cases.

Issue 2 arrises as there are two implementations of batch norm in keras (fused/unfused). Tensors of rank 5 are not supported in the faster fused version, so a check is performed based on the input shape and the correct version should be selected. For some reason this check isn't working on the macOS arm chipsets (possibly a tensorflow-metal or keras version issue) and this causes the shape mismatch at inference.

Checking behaviour on an Ubuntu installation the batch norm does seem to default to the non-fused version so explicitly setting fused=False at network construction should avoid the issue altogether.

This commit applies the fix to mri_synthseg, mri_synthsr and mri_segment_thalamic_nuclei_dti_cnn but the issue is likely still present in other tensorflow based tools.

### DIFF
--- a/mri_easyreg/mri_easyreg
+++ b/mri_easyreg/mri_easyreg
@@ -412,7 +412,7 @@ def reformat_to_list(var, length=None, load_as_numpy=False, dtype=None):
     if var is None:
         return None
     var = load_array_if_path(var, load_as_numpy=load_as_numpy)
-    if isinstance(var, (int, float, np.int8, np.int16, np.int32, np.int64, np.float16, np.float32, np.float64, np.float128)):
+    if isinstance(var, (int, float, np.int8, np.int16, np.int32, np.int64, np.float16, np.float32, np.float64)):
         var = [var]
     elif isinstance(var, tuple):
         var = list(var)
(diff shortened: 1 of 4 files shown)


htregidgo:
Thanks for merging these changes. From a quick grep on BatchNormalisation the following files may also be affected:

1. cnn_sphere_register/ext/neuron/neuron/models.py
2. deeplearn_utils/unet_model.py
3. mri_sclimbic_seg/mri_sclimbic_seg
4. mri_segment_hypothalamic_subunits/mri_segment_hypothalamic_subunits
5. mri_synthsr/mri_synthsr_hyperfine
6. psacnn_brain_segmentation/psacnn_brain_segmentation/deeplearn_utils/unet_model.py
7. python/fsdeeplearn/models.py
8. python/fsdeeplearn/scripts/train_ev_semi.py
9. recon_all_clinical/python/mri_synth_surf.py
10. sscnn_skullstripping/sscnn_skullstripping/deeplearn_utils/unet_model.py

PR #1198, opened 2024-04-23
nf: synthseg/easyreg autocrop

htregidgo:
In cases where the Field of View (FOV) is set too large on low resolution volumes, resampling to 1mm iso in synthseg gives  extremely large tensors (350+ to a side). In such cases the system can struggle to find large enough contiguous blocks of memory leading to 'bad alloc' errors. In such cases much of the FOV is unneeded and so can be cropped out to reduce memory pressure and compute time. 

Cropping manually with the --crop option requires knowledge of head placement and dimension and is limited to the centre of the FOV. However, synthseg already re-normalises input images and zero's out background noise. Here we've added the --autocrop flag to allow automatic cropping to the smallest bounding box containing everything that is not zeroed out by the normalisation step. 

Additional fixes include:
1. Multi-channel support for resampling using mri_easywarp
2. Handling of numerical roundoff when matching cortical parcellation labels in mri_easyreg

### DIFF
--- a/mri_easyreg/mri_easyreg
+++ b/mri_easyreg/mri_easyreg
@@ -103,8 +103,8 @@ def main():
         print('   Reading reference image')
         ref_image, ref_aff, ref_h, ref_im_res, ref_shape, ref_pad_idx, ref_crop_idx = preprocess(path_image=args.ref,
                                                                                                  crop=None, min_pad=128,
-                                                                                                 autocrop=args.autocrop,
-                                                                                                 path_resample=None)
+                                                                                                 path_resample=None,
+                                                                                                 autocrop=args.autocrop)
         print('   Setting up segmentation net')
         segmentation_net = build_seg_model(model_file_segmentation=path_model_segmentation,
                                            model_file_parcellation=path_model_parcellation,
@@ -140,8 +140,8 @@ def main():
         print('   Reading floating image')
         flo_image, flo_aff, flo_h, flo_im_res, flo_shape, ref_pad_idx, ref_crop_idx = preprocess(path_image=args.flo,
                                                                                                  crop=None, min_pad=128,
-                                                                                                 autocrop=args.autocrop,
-                                                                                                 path_resample=None)
+                                                                                                 path_resample=None,
+                                                                                                 autocrop=args.autocrop)
         if segmentation_net is None:
             print('   Setting up segmentation net')
             segmentation_net = build_seg_model(model_file_segmentation=path_model_segmentation,
@@ -503,7 +503,7 @@ def load_volume(path_volume, im_only=True, squeeze=True, dtype=None, aff_ref=Non
 
 
 
-def preprocess(path_image, n_levels=5, crop=None, autocrop=False, min_pad=None, path_resample=None):
+def preprocess(path_image, n_levels=5, crop=None, min_pad=None, path_resample=None, autocrop=False):
     # read image and corresponding info
     im, _, aff, n_dims, n_channels, h, im_res = get_volume_info(path_image, True)
     if n_dims < 3:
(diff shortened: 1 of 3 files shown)


jnolan14 (commented):
This looks pretty good to me, just 2 comments:
1. In the functions/methods where the 'autocrop' arg is added, it might be best to stick that new arg at the end of the signature, so any preexisting calls to those methods that rely on positional args will not break
2. We want to avoid code duplication of the autocrop functionality, could that be accomplished by calling the 'crop_to_bbox' method in surfa that's a member of the FramedArray class? If not, then maybe we want to move that functionality into the surfa utils, so there's only one instance of it in the code base.

htregidgo:
> 1. In the functions/methods where the 'autocrop' arg is added, it might be best to stick that new arg at the end of the signature, so any preexisting calls to those methods that rely on positional args will not break

Good catch! I've made the change.

> 2. We want to avoid code duplication of the autocrop functionality, could that be accomplished by calling the 'crop_to_bbox' method in surfa that's a member of the FramedArray class? If not, then maybe we want to move that functionality into the surfa utils, so there's only one instance of it in the code base.

In principle yes but theres a few things to keep in mind:
1. synthseg/easyreg image read is handled directly by nibabel not surfa so everything is np.arrays not FramedImages
 a. (there are some sf.load_volumes but they are only used for things like hacking together resampling to example geometries [see line 1477](https://github.com/htregidgo/freesurfer/blob/5e624a8237e803540e7c5f7483b44f5fc61f0c01/mri_synthseg/mri_synthseg#L1477))
2. we need to retain the cropping indices, which aren't returned by sf.image.FramedImage.crop_to_bbox, as they're used later to insert the segmentation in the correct location for output
 a. (possibly wouldn't be an issue if synthseg was completely refactored to use surfa rather than nibabel but that is definitely a much larger code change)
3. crop_to_bbox allows a margin to be specified around the content bbox, but in the case of autocrop this needs to be dynamically calculated so that the dimensions of the array are divisible by the minimum patch size of the network (in this case 32) otherwise prediction will throw an error.
4. mri_synthseg, mri_synthsr, mri_easyreg, mri_segment_hypothalamic_subunits, mri_segment_thalamic_nuclei_dti_cnn and probably others are already riddled with code duplication with large chunks of ctrl-c+v code that has minor modifications dotted around at random. In fact the majority of the code in these scripts are duplicate functions. A refactor needs to happen at some point but it is a pretty big job.
 a. for example vastly similar crop_volume code is [here](https://github.com/htregidgo/freesurfer/blob/5e624a8237e803540e7c5f7483b44f5fc61f0c01/mri_synthseg/mri_synthseg#L1965), [here](https://github.com/htregidgo/freesurfer/blob/5e624a8237e803540e7c5f7483b44f5fc61f0c01/mri_easyreg/mri_easyreg#L1436), [here](https://github.com/htregidgo/freesurfer/blob/5e624a8237e803540e7c5f7483b44f5fc61f0c01/mri_segment_hypothalamic_subunits/mri_segment_hypothalamic_subunits#L1406) and [here](https://github.com/htregidgo/freesurfer/blob/5e624a8237e803540e7c5f7483b44f5fc61f0c01/mri_segment_thalamic_nuclei_dti_cnn/mri_segment_thalamic_nuclei_dti_cnn#L1788)

I'm aware leaving the autocrop functionality out of surfa is just kicking the problem down the road but moving it into utils requires making some decisions.

jnolan14:
Method signatures look good to me!

> 1. synthseg/easyreg image read is handled directly by nibabel not surfa so everything is np.arrays not FramedImages

That's a good point, I hadn't realized that. Although we could convert them to sf.FramedArray objs, then do the cropping. If we need to save the crop bounds, then we can get that by calling FramedArray.bbox(), which is called under the hood by crop_to_bbox (maybe we should have crop_to_bbox return the crop bounds by default?). I think what you mention in point 4, that this code duplication already exists, makes the surfa stuff I just mentioned a largely a moot point.

I'd say if that code wasn't already copy pasted it would be worth refactoring it and stuffing it in the surfa.utils now, but the duplication is already there.
If this looks good to Avnish, I'll go ahead and merge!

avnishks (approved):

jnolan14 (approved):
Merging!

PR #1201, opened 2024-04-26
Update registration.py to perform just landmark detection

jeiglesias:


jnolan14:
Jon said this is working

PR #1202, opened 2024-04-26
Create func_fiducials_detection.py for detecting fiducials

jeiglesias:


jnolan14:
Jon said this is working

PR #1220, opened 2024-07-17
IMPORTANT FIX PARCELLATION VOLUMES

BBillot:
there was a mistake for the normalisation of the cortical volumes when doing parcellation.
I also included some housekeeping for synthseg and friends

### DIFF
--- a/mri_segment_hypothalamic_subunits/mri_segment_hypothalamic_subunits
+++ b/mri_segment_hypothalamic_subunits/mri_segment_hypothalamic_subunits
@@ -565,29 +565,26 @@ def postprocess(prediction, im_shape, crop_idx, labels, aff):
 
 
 def write_csv_file(volumes, filename, subject, write_header, open_type):
-    '''
-    Writes volumetric stats to csv file.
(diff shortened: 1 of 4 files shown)


mri_segment_hypothalamic_subunits/mri_segment_hypothalamic_subunits
```diff
@@ -1234,14 +1231,15 @@ def list_subfolders(path_dir, whole_path=True, expr=None, cond_type='or'):
 
 def mkdir(path_dir):
```
avnishks, line 1232: A more pythonic way of making recursive directories would be to use:

def mkdir(path_dir):
    os.makedirs(path_dir, exist_ok=True)

BBillot: yes I agree, but this function is from a long time ago, and using os.makedirs consistently across all the repos would take me quite some time.
The main thing for this PR is really to fix a problem with the cortical parcellation volumes, so it's quite urgent. The sooner the better :)
Thanks !
Benjamin

PR #1233, opened 2024-10-02
AANSegment-shell-update

markolchanyi:
Added missing preprint citation information and segmentation volume viewing instructions to SegmentNuclei shell script in AANSegment


jnolan14:
Looks good to me. 
We may want to add functionality to the matlab exes or the bash wrapper to move/reformat the output *.txt under mri/ to be *.stats files under stats/, so the concatenation util will work for this as well.

PR #1234, opened 2024-10-10
moved AANSegment volumes txt file to /stats/.stats

markolchanyi:
Added line to move the arousalNetworkVolumes.${SUFFIX}.txt outputted in /mri over to the /stats directory as arousalNetworkVolumes.${SUFFIX}.stats. 

mkdir -p .../stats/... is there just in-case one runs the script without recon-all. 

The name arousalNetworkVolumes is hard-coded since its name is created in downstream code. I will change this when we convert to python.

### DIFF
--- a/AANsegment/SegmentAAN.sh
+++ b/AANsegment/SegmentAAN.sh
@@ -231,6 +231,9 @@ else
 endif
 echo $returnVal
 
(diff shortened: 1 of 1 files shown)


jnolan14:
Looks good to me. Do you have a a few test cases I can use to validate the Concat..results.sh script?

PR #1245, opened 2024-11-07
Create segment_fast.py for fast segmentation

jeiglesias:


jnolan14:
This looks ok, but there's a lot of path manipulation going on at the top of the script. Would it be better to handle those things in how fspython is configured, or adding the required paths to PYTHONPATH in the wrapper script that calls this via fspython?

PR #1247, opened 2024-11-07
Replaced AANSegment Matlab executable with Matlab8.4 CentOS7 executable 

markolchanyi:
Replaced old AANSegment Matlab src compilation with Matlab8.4 CentOS7 compilation in AANSegment to resolve symbol issues.


jnolan14:
Tested and confirmed that the required libs could be found on Centos7 systems

PR #1248, opened 2024-11-11
Update mri_easyreg to write out affine matrices

jeiglesias:


jnolan14:
The value of 'combinations' on line 1660 is undefined. What is this supposed to be, it looks like it's supposed to be a function call?

PR #1277, opened 2025-01-30
updated csvprint to python3

jnolan14:
Updated shebang and syntax to python 3.
Removed unused imports and semicolons.
Changed file read to read non-binary files.

Does this need to support files in both binary and non-binary format?


m-reuter:
probably csv is ascii, so I doubt that binary read was needed.

wxyanggz:
Hello, after I updated csvprint, the value of sTIV is still 0. 
Moreover, what is mentioned in https://ftp.nmr.mgh.harvard.edu/pub/dist/lcnpublic/dist/csvprint_8.0.0_patch/README.md is that the value of eTIV is 0. 
How should I solve the problem that the value of sTIV is 0?

jnolan14:
That is a typo in the patch notes, this should fix the issue with the 0 value reported for sTIV.
However, if the file containing the stats exists when recon-all is rerun, that file will not be updated. Can you try deleting the file and rerunning?

PR #1278, opened 2025-01-30
BF: Fixed undefined refs, updated shebang

jnolan14:
Fixed undefined references:
- xrange -> range, looks like this was hanging around from python2
- basedir -> subjectid

Both undefined references were in sections that would only be reached if an optional arg was passed, or a specified subject didn't have the proper number of timepoints.


m-reuter (approved):
Looks good to me. thanks for the fixes

PR #1279, opened 2025-01-30
BF: fixed undefined refs in sections handling errors. 

jnolan14:
Dropped appending timepoint to list of error cases, as it is was undefined when that error is handled, and the bad timepoint never seems to be reported to the user when appended to that list.


m-reuter (approved):
Thanks for fixing. This makes sense.

PR #1280, opened 2025-01-30
BF: merge_stats_tables

jnolan14:
Changed the undefined error 'BadFileError' to be 'FileNotFoundError', and changed the try/catch block to catch FileNotFoundError. 
Removed unused imports
Updated shebang to python3


m-reuter (approved):
looks good

PR #1296, opened 2025-03-30
fix #1294: check if output matrix is large enough to hold a copy

### LINKED ISSUE #1294: Bug in utils/matrix.cpp: MatrixCopy Doesn't check if output matrix is large enough to hold a copy
Consider the following code

```cpp
MATRIX *m1 = MatrixAlloc(3, 3, MATRIX_REAL);
MATRIX *m2 = MatrixAlloc(2, 2, MATRIX_REAL);
MatrixConstVal(5, 3, 3, m1);
  
MatrixCopy(m1, m2);
```
the above code causes a segmentation fault, because `MatrixCopy` didn't check
if `m2` is large enough, this only happens when passing an output array

this code works however:

```cpp
MATRIX *m2 = MatrixCopy(m1, nullptr);
```

aboualiaa:
also initialize it to zero just in case it's larger than input matrix

### DIFF
--- a/utils/matrix.cpp
+++ b/utils/matrix.cpp
@@ -161,6 +161,8 @@ MATRIX *MatrixCopy(const MATRIX *mIn, MATRIX *mOut)
   cols = mIn->cols;
 
   if (!mOut) mOut = MatrixAlloc(rows, cols, mIn->type);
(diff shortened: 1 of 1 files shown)


yhuang43:
I made a mistake editing your changes. It is fix in commit 6a3e7c4

PR #1298, opened 2025-03-30
fix #1297: OpenMatrixDeterminant in utils/numerics.cpp returns ERROR_BAD_PARAM on non square matrices

### LINKED ISSUE #1297: bug in utils/numerics.cpp: OpenMatrixDeterminant returns 0 for non-square matrices
This conflicts with the definition of a determinant per se, it should return an error value

aboualiaa:


aboualiaa:
closed pull request because it doesn't take into account negative determinants, will try to come up with another solution

PR #1299, opened 2025-03-31
Enhancements for mri_synthstrip Docker builds

gdevenyi:
GPU Dockerfile
- switch to minimal python base image (~200MB savings)
- add support for docker build cache of pip files
- use github tarball of hash to avoid git install
- use official pytorch wheel

New CPU Dockerfile
- features as above
- 3.79GB (!!!) savings

I'd like to request the CPU tag variant be published alongside the regular one in Dockerhub.

Ref:
https://pytorch.org/get-started/previous-versions/

### DIFF
--- a/mri_synthstrip/Dockerfile.cpu
+++ b/mri_synthstrip/Dockerfile.cpu
@@ -1,0 +1,33 @@
+# base image with Python and SynthStrip on path (eventually)
+FROM python:3.10-slim AS base
+ENV FREESURFER_HOME="/freesurfer"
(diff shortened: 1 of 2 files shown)


gdevenyi:
I had to update the HASH for the surfa package as it was crashing on an type error for `int_t`

mu40:
This is great, thank you! We'll look into publishing a CPU-only version.

gdevenyi:
I haven't used the other ML tools in freesurfer, but if there are similar dockerfiles they may benefit from the same treatment.

PR #1304, opened 2025-04-17
change sscnn mentions to synthstrip

Ash-Bal:
Cleaning up some paths. We don't use sscnn anymore and have transitioned to synthstrip so just reflecting the change in the file structure


jnolan14 (approved):

jnolan14:
Now that sscnn has been replaced and the names updated in the current infant install, it should now be safe to drop the top level sscnn_skullstripping dir from the top level of the repo

PR #1323, opened 2025-05-29
Add procps to containers to enable nextflow use

gdevenyi:
nextflow's container management uses ps to monitor container processes so adding this in to enable it. Minimal impact on the container.

### DIFF
--- a/mri_synthstrip/Dockerfile.cpu
+++ b/mri_synthstrip/Dockerfile.cpu
@@ -4,6 +4,12 @@ ENV FREESURFER_HOME="/freesurfer"
 ENV PYTHONUSERBASE="$FREESURFER_HOME/env"
 ENV PATH="$FREESURFER_HOME:$PATH"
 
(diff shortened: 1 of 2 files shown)


mu40:
If this is still relevant, have you tested if the `procps` tools are actually present in the image? Since they're added in an intermediate build stage, I wouldn't expect them to make it into the final stage, which derives from `base`.

gdevenyi:
> If this is still relevant, have you tested if the `procps` tools are actually present in the image? Since they're added in an intermediate build stage, I wouldn't expect them to make it into the final stage, which derives from `base`.

My bad I indeed put this in the wrong place. Now added in the base container instead.

mu40:
I suspect the impact on the image is still minimal. Could you add a one-line comment to your run instruction noting who needs `procps` and why?

gdevenyi:
Comment added :+1:

Issue #1335, opened 2025-06-30
recon-all -autorecon1 fails on v8.0.0


abroy77:
## Example command:

``` 
mksubjdir sub_1
recon-all -s sub_1 -autorecon1 -i T1.nii.gz 
```
## Error message: 
```
#@# CC Seg Mon Jun 30 11:39:58 UTC 2025

 seg2cc --s sub_1 

ERROR: cannot find /execute/subjects/sub_1/mri/norm.mgz
Linux 61355d480d39 6.11.0-26-generic #26~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Apr 17 19:20:47 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

```

freesurfer version (v8.0.0)
```
freesurfer-linux-ubuntu22_x86_64-8.0.0-20250204-0f2bd3a
```
## Notes
It seems the `v8.0.0` `autorecon1` runs `seg2cc` which requires `norm.mgz` as its input.

## Workaround
pass the `-gcareg` and `-canorm` options to the `recon-all` command. eg:
```
recon-all -s sub_1 -autorecon1 -gcareg -canorm -i T1.nii.gz 
```
## Fix:
I will make a PR which updates the `-autorecon1` flag to add `canorm` and `gcareg` as default so the user does not need to provide the additional flags.

## Logs

[recon-all.log](https://github.com/user-attachments/files/20979404/recon-all.log)
[recon-all-status.log](https://github.com/user-attachments/files/20979405/recon-all-status.log)
[build-stamp.txt](https://github.com/user-attachments/files/20979403/build-stamp.txt)

abroy77:
I suggest this be added to the `Known Issues` page for `v8.0.0`

yhuang43:
I think the issue has been addressed in commit 6cb7d84

PR #1336, opened 2025-06-30
Fix recon-all '-autorecon1' flag

abroy77:
Make the autorecon1 flag do gca_reg and ca_norm as they are needed for seg2cc which is a later step in the `autorecon1` pipeline. Alternatively the inputs to seg2cc may need to use a file other than `norm.mgz` for the `autorecon1`  pipeline to work.
Related #1335


yhuang43:
I think the issue has been addressed in commit 6cb7d84

PR #1349, opened 2025-08-06
recon-all-clinical script error

arnaudbore:
recon-all-clinical.sh was not working properly and I was getting this error:

`then: then/endif not found.`

This quick fix seems to work


kharitz (approved):
Looks good.

PR #1351, opened 2025-08-11
quick fix in mri_histo_util/ERC_bayesian_segmentation/ for NextBrain

chourroutm:


chourroutm:
@jeiglesias in case you want to have a look at these

jeiglesias (approved):
we should accept this one

jeiglesias:
> @jeiglesias in case you want to have a look at these

Thanks for these!
-GMM: yes! 
-BF: the corrected image is already written when analyzing the left hemi

yhuang43:
@chourroutm @jeiglesias I can't seem to find mri_histo_util/mri_histo_atlas_segment. I see these files in the directory:

mri_histo_util/mri_histo_atlas_segment_fireants
mri_histo_util/mri_histo_atlas_segment_fullbayesian
mri_histo_util/mri_histo_atlas_segment_synthmorph

jeiglesias:
There isn't such a command, please see:
https://github.com/freesurfer/freesurfer/blob/dev/mri_histo_util/README.md

yhuang43:
@jeiglesias @chourroutm @jnolan14
I'm little confused. What should we do with mri_histo_util/mri_histo_atlas_segment then? Has the file been renamed?

jnolan14:
I'll let @jeiglesias confirm or deny this, but I think everything except for the mri_histo_atlas_segment_fireants and its associated wrapper and dependencies can be dropped as that's the newest/fastest implementation that they have working.

chourroutm:
Indeed, this is no longer relevant as the standard and the "fast" versions have been removed.

jeiglesias:
yeah it was too messy to keep all versions. This is the one we're pushing https://www.biorxiv.org/content/10.1101/2025.09.22.673638v1

PR #1353, opened 2025-08-16
Update SynthSR 

### LINKED ISSUE #1352: Update SynthSR for modern Python
SynthSR is a terrific tool. However, the [original code requires Python 3.6](https://github.com/BBillot/SynthSR/blob/9e63935503d97a433fcf7082b9f69abc5e7a1ddd/setup.py#L8) and the version with FreeSurfer 8.1 is limited to Python 3.8, both of which are [end of life](https://devguide.python.org/versions/). This week [Tensorflow was upgraded to support Python 3.13](https://pypi.org/project/tensorflow/). 

I have updated [SynthSR](https://github.com/neurolabusc/py_synthsr) to use modern tensorflow, support versions of Python up to 3.13, and remove unused dependencies (surfa). It would be great if these changes could be merged into FreeSurfer. Happy to make a minimal PR.

In my testing, the resulting uint8 images created by modern versions of tensorflow have over 99% of voxels identical with the old versions, with no voxel differing by more than 1. This suggests changes in rounding behavior typical of software upgrades (e.g. fused multiply–add optimizations are not only faster but have higher precision).

neurolabusc:
- closes #1352
 
 This is the minimal change to SynthSR
 - Detects the TensorFlow version to ensure forward and backward compatibility with the removed fused option.
 - Removes the `surfa` dependency.
 - If `FREESURFER_HOME` is not set, tests if `models` are located in the same folder as the script.

### DIFF
--- a/mri_synthsr/mri_synthsr
+++ b/mri_synthsr/mri_synthsr
@@ -7,7 +7,6 @@ import time
 import numpy as np
 import argparse
 import nibabel as nib
-import surfa as sf
 from datetime import timedelta
 from scipy.ndimage import gaussian_filter
 from scipy.interpolate import RegularGridInterpolator
@@ -24,6 +23,15 @@ backend.set_image_data_format('channels_last')
 # ================================================================================================
 #                                         Main Entrypoint
 # ================================================================================================
+
+def batch_normalization(axis, name, tensor):
+    major, minor, patch = (int(x) for x in tf.__version__.split(".")[:3])
+    if (major, minor) <= (2, 13):
+        # Old behavior requires fused=False
+        return keras.layers.BatchNormalization(axis=axis, name=name, fused=False)(tensor)
+    else:
+        # New behavior (fused argument removed)
+        return keras.layers.BatchNormalization(axis=axis, name=name)(tensor)
 
 
 def main():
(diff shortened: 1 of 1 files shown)


yhuang43:
Hi Chris,

The PR looks good! Just couple of things that I noticed:

1. To preserve the script's behavior and make it less confusing for Freesurfer users, maybe it should use 'importlib.util.find_spec("py_synthsr")' to determine if 'py_synthsr' is importable. If it is not importable, just 'raise RuntimeError('FREESURFER_HOME is not set. Please source freesurfer.')'

2. for postprocessing, the line 'pred = postprocess(pred, pad_idx, aff, disable_sharpening)' is missing clip(0, 128) operation. 
    It should be 'pred = postprocess(pred.clip(0,128), pad_idx, aff, disable_sharpening)'.

4. line 622 - 625, I'm not sure if 'last_tensor' will be a list.  Can we change line 625 'input_shape = list(last_tensor.shape)[1:]' to 'input_shape = last_tensor.shape.as_list()[1:]'?

neurolabusc:
Sounds good. Please feel free to make the edits you recommend.

yhuang43:
I did the edits. Can you take a look before I merge the changes?

mri_synthsr/mri_synthsr
```diff
@@ -621,8 +620,7 @@ def conv_dec(nb_features,
         last_tensor = input_model.output
         if isinstance(last_tensor, list):
             last_tensor = last_tensor[0]
-        #CR input_shape = last_tensor.shape.as_list()[1:]
-        input_shape = list(last_tensor.shape)[1:]
+        input_shape = last_tensor.shape.as_list()[1:]
```
neurolabusc: @yhuang43 have you tested this on a modern version of TensorFlow? I believe you have to adopt my change to have this be forward compatible.

yhuang43: No, I didn't test my changes on newer version of TensorFlow. I did test your changes in Freesurfer. So, I rolled back the changes I made. Let me know the changes still work for you.

yhuang43: @neurolabusc What are the python, numpy, tensorflow versions in your environment?

neurolabusc:
@yhuang43 you can try my patch out independently of FreeSurfer with [this modern Python repository](https://github.com/neurolabusc/py_synthsr). I am not sure what Python version FreeSurfer is targeting, but the challenge I faced was creating a single script that supported the EOL version of Python provided with FreeSurfer as well as the latest versions supported by TensorFlow. I would suggest the best way to evaluate my modifications is to use the same script with the Python used with your FreeSurfer development as well as in an isolated Python environment with the latest release of tensorflow. At the moment, [TensorFlow 2.20.0](https://www.tensorflow.org/install/source#tested_build_configurations) supports Python 3.9-3.13, so this is reflected in the [my repositories requirements](https://github.com/neurolabusc/py_synthsr/blob/04ab5548f4609c2b44ca51b3f4bc319b585dafa7/pyproject.toml#L13). I realize that Python 3.14 has been released, tensorflow is the rate limiting factor.

So the trick was to create a script that supports both the version of Python used with FreeSurfer as well as recent releases. This was the case last August when I submitted my PR. It covered the Python version from that time as well as the most recent Python supported by tensor flow at that time (and these latter ranges have not changed since then). Therefore, assuming the version of Python for FreeSurfer is similar or more recent now, I suspect my changes remain effective.

yhuang43:
@neurolabusc Thanks for the input. I did test your PR inside Freesurfer environment. I wanted to make sure my following modifications still work for you
image

neurolabusc:
Can I suggest you test your changes yourself with my isolated Python repository. That will close the loop for you to evaluate the changes.

yhuang43:
I tested it in py_synthsr environment. I confirm your changes work. In tensorflow 2.20.0, 'last_tensor=input_model.output' returns a list, and tensor.shape.as_list() doesn't work for unknown shape.

PR #1372, opened 2025-09-24
fixed and reorganized NextBrain

jeiglesias:


jnolan14:
Might be able to simplify some logic with the argument validation, but will should all run fine

PR #1383, opened 2025-10-20
Update inference.py to write label names in csv file with ROI volumes

jeiglesias:


jnolan14:
Functionally this should work, but we should address 2 things:

1. The sys.path.append call should be unnecessary, as that path is added to PYTHONPATH via the fspython bash wrapper script.
2. Is there any reason not to use the cLUT functionality in surfa? The default cLUT can be loaded with sf.freesurfer.labels() and returns an ordered dict of labels that can be similarly used.

PR #1406, opened 2025-12-11
nf: Pass GPU setting to mri_synthstrip in recon-all

MegaByte:
Similar to the existing option for mri_synthseg, this activates the GPU option in mri_synthstrip when the -gpu flag is set in recon-all.


yhuang43:
The proposed change is included in commit 856d584.

PR #1430, opened 2026-03-02
Changes needed to add mri_tumorsynth

fprados:
I've added the tumorsynth script and needed to upgrade the nnunetv1 python to 3.9


fprados:
I've done an extra change then I open a new pull request

Issue #1432, opened 2026-03-22
recon-all crash in v8.2.0 when using expert options


s-goulis:
Hello FreeSurfer developers,

I encountered two reproducible bugs in the latest update, Stable 8.2.0 release (Mar 18 2026), when running recon-all with an expert options file on a 0.7 mm 7T MP2RAGE dataset (-hires flag used, run on WSL2/Ubuntu 22.04).

### **Issue 1**
recon-all crashes after mris_sphere because it evaluates if($XOptsFile), which expands to the expert-options path and triggers tcsh’s `if: Expression Syntax.` error.

**Suggested fix**
Replace:
```
if($XOptsFile)
```
with 
```
if($#XOptsFile):
```

### **Issue 2**
After patching the first issue locally, recon-all proceeds further but then passes --expert to rca-surfreg. However, rca-surfreg currently recognizes -expert but not --expert, which causes:
`ERROR: Flag --expert unrecognized.
`
**Suggested fix**
In rca-surfreg, add support for --expert, e.g.:
```
case "--expert":
case "-expert":
```

These issues were reproduced with an expert options file containing only `mris_inflate -n 60`. I will also submit a pull request with proposed fixes for both issues. Thank you for continuing to develop and maintain FreeSurfer.

Best regards,
Sophocles Goulis

buildqa:
Please see the recent email on the freesurfer help list notifying users that some updates (including this issue) are now available via following the instructions at, https://surfer.nmr.mgh.harvard.edu/fswiki/rel7downloads/rel8notes

buildqa:
closing since no activity in the past 2 months

PR #1433, opened 2026-03-22
Fix expert option handling in v8.2.0 recon-all and rca-surfreg

s-goulis:
This PR fixes two expert-options issues:

1. recon-all evaluates` if($XOptsFile)`, which can expand to a path and trigger tcsh "`if: Expression Syntax.`"
2. rca-surfreg does not recognize` --expert`, although recon-all passes that flag.

Changes:
- replace `if($XOptsFile)` with `if($#XOptsFile)` in recon-all
- add support for `--expert` alongside `-expert` in rca-surfreg

Fixes Issue #1432

### DIFF
--- a/scripts/rca-surfreg
+++ b/scripts/rca-surfreg
@@ -248,6 +248,7 @@ while( $#argv != 0 )
       set RunIt = 0;
       breaksw
 
+    case "--expert":
     case "-expert":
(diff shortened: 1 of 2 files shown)


yhuang43:
The fix is implemented in commit 0068e66

